PR #2, opened 2017-04-11
Add code that gives an nice Array interface to a vector of arrays

gabrielgellner:
Let me know what you need. My gitfu is weak.

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -1,65 +1,49 @@
-abstract AbstractVectorOfArray{T, N} <: AbstractArray{T, N}
-
 # Based on code from M. Bauman Stackexchange answer + Gitter discussion
-type VectorOfArray{T, N, A} <: AbstractVectorOfArray{T, N}
+
+type VectorOfArray{T, N, A} <: AbstractArray{T, N}
   data::A # A <: AbstractVector{<: AbstractArray{T, N - 1}}
+  #TODO: I don't really use dims, how do I get rid of it, and still have everything work?
   dims::NTuple{N, Int}
-  ragged::Bool
 end
 
 function VectorOfArray(vec::AbstractVector)
-  # Only allow a vector of equally shaped subtypes, this will not work for ragged arrays
-  ragged = !all(size(vec[1]) == size(v) for v in vec)
-  VectorOfArray(vec, (size(vec[1])..., length(vec)), ragged)
+    # Only allow a vector of equally shaped subtypes, this will not work for ragged arrays
+    #@assert all(size(vec[1]) == size(v) for v in vec)
+    VectorOfArray(vec, (size(vec[1])..., length(vec)))
 end
 
-VectorOfArray{T, N}(vec::AbstractVector{T}, dims::NTuple{N}, ragged::Bool) = VectorOfArray{eltype(T), N, typeof(vec)}(vec, dims, ragged)
-@inline function Base.size(S::AbstractVectorOfArray)
-  if S.ragged
-    return size(S.data)
-  else
-    return S.dims
-  end
-end
+VectorOfArray{T, N}(vec::AbstractVector{T}, dims::NTuple{N}) = VectorOfArray{eltype(T), N, typeof(vec)}(vec, dims)
+Base.size(S::VectorOfArray) = (size(S.data[1])..., length(S.data))
 
-@inline function Base.getindex{T, N}(S::AbstractVectorOfArray{T, N}, I::Vararg{Int, N})
-  @boundscheck checkbounds(S, I...)
-  S.ragged && throw(BoundsError("A ragged VectorOfArray does not support Cartesian indexing"))
-  S.data[I[end]][Base.front(I)...]
+@inline function Base.getindex{T, N}(S::VectorOfArray{T, N}, I::Vararg{Int, N})
+    @boundscheck checkbounds(S, I...) # is this needed?
+    S.data[I[end]][Base.front(I)...]
 end
 # Linear indexing will be over the container elements, not the individual elements
 # unlike an true AbstractArray
-@inline Base.getindex{T, N}(S::AbstractVectorOfArray{T, N}, I::Union{Int, Colon, AbstractArray{Int}}) = S.data[I]
+@inline Base.getindex{T, N}(S::VectorOfArray{T, N}, I::Int) = S.data[I]
+@inline Base.getindex{T, N}(S::VectorOfArray{T, N}, I::AbstractArray{Int}) = S.data[I]
 
-Base.copy(S::AbstractVectorOfArray) = VectorOfArray(copy(S.data), S.dims, S.ragged)
+Base.copy(S::VectorOfArray) = VectorOfArray(copy(S.data), S.dims)
 
-Base.sizehint!{T, N}(S::AbstractVectorOfArray{T, N}, i) = sizehint!(S.data, i)
+Base.sizehint!{T, N}(S::VectorOfArray{T, N}, i) = sizehint!(S.data, i)
 
-function Base.push!{T, N}(S::AbstractVectorOfArray{T, N}, new_item::AbstractArray)
-  if S.dims[1:(end - 1)] == size(new_item)
-    S.dims = (S.dims[1:(end - 1)]..., S.dims[end] + 1)
-  else
-    S.ragged = true
-    # just revert down to a vector of elements, make the dims data meaningless
-    #TODO: this is stupid ugly
-    S.dims = tuple(-ones(Int, N)...)
-  end
-  push!(S.data, copy(new_item))
+function Base.push!{T, N}(S::VectorOfArray{T, N}, new_item::AbstractVector)
+    #@assert S.dims[1:(end - 1)] == size(new_item)
+    #S.dims = (S.dims[1:(end - 1)]..., S.dims[end] + 1)
+    push!(S.data, new_item)
 end
-function Base.append!{T, N}(S::AbstractVectorOfArray{T, N}, new_item::AbstractVectorOfArray{T, N})
+
+function Base.append!{T, N}(S::VectorOfArray{T, N}, new_item::VectorOfArray{T, N})
     for item in copy(new_item)
-      push!(S, item)
+        push!(S, item)
     end
     return S
 end
 
+
 # The iterator will be over the subarrays of the container, not the individual elements
 # unlike an true AbstractArray
-Base.start{T, N}(S::AbstractVectorOfArray{T, N}) = 1
-Base.next{T, N}(S::AbstractVectorOfArray{T, N}, state) = (S[state], state + 1)
-Base.done{T, N}(S::AbstractVectorOfArray{T, N}, state) = state >= length(S.data) + 1
-
-# convert to a regular, sense array
-function vecarr_to_arr(va::AbstractVectorOfArray)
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
Would it be difficult to check what the cost of the ragged flag is?

What happens if you just put this on a `Vector{Float64}`? Curious. I don't think we'll need this, but I wonder what it does.

gabrielgellner:
Yeah I wanted to benchmark this. Do you have any examples of how I might best do this? Should I just make a couple of different sized arrays and check out the overhead of doing indexing (the part that scares me the most)?

src/vector_of_array.jl
```diff
@@ -0,0 +1,65 @@
...
+# Linear indexing will be over the container elements, not the individual elements
+# unlike an true AbstractArray
+@inline Base.getindex{T, N}(S::AbstractVectorOfArray{T, N}, I::Union{Int, Colon, AbstractArray{Int}}) = S.data[I]
+
+Base.copy(S::AbstractVectorOfArray) = VectorOfArray(copy(S.data), S.dims, S.ragged)
+
+Base.sizehint!{T, N}(S::AbstractVectorOfArray{T, N}, i) = sizehint!(S.data, i)
+
+function Base.push!{T, N}(S::AbstractVectorOfArray{T, N}, new_item::AbstractArray)
+  if S.dims[1:(end - 1)] == size(new_item)
```
ChrisRackauckas: `if S.ragged && S.dims[1] == size(new_item)`

ChrisRackauckas: wait actually, this is for not ragged? I don't quite understand what's going on here.

src/vector_of_array.jl
```diff
@@ -0,0 +1,65 @@
+abstract AbstractVectorOfArray{T, N} <: AbstractArray{T, N}
+
+# Based on code from M. Bauman Stackexchange answer + Gitter discussion
+type VectorOfArray{T, N, A} <: AbstractVectorOfArray{T, N}
+  data::A # A <: AbstractVector{<: AbstractArray{T, N - 1}}
+  dims::NTuple{N, Int}
```
ChrisRackauckas: Why is this a tuple if it's changing with each push?

ChrisRackauckas: Would it be better as a `Vector{Int}`?

gabrielgellner: Yeah I will make that change, I think I originally did this just to make it more Array like, but it clearly doesn't make sense as it is not static.

src/vector_of_array.jl
gabrielgellner: We are updating the dims attribute if it matches the current shape of the other elements (we drop the last element as this is the length of the containing vector). Otherwise we make it ragged. But you show a bug, I need to check that it has already become ragged in which case I don't need to do anymore checking. Will fix this up.

src/vector_of_array.jl
gabrielgellner: The only issue is can I make the field readonly? Might cause strange issues if the user mutates the dim field.

ChrisRackauckas: Don't worry about that. If a user does that, they deserve problems.

ChrisRackauckas:
It would be good to know the numbers for the overhead of indexing, but since there really isn't much we can do about it (by design) it's moreso for understanding when someone should convert to a matrix/tensor.

There are two things here. Speed of building and speed of indexing. Speed of building is first. I plan to swap out the current array that's used in each solver with a `VectorOfArray` (currently it's a `Vector{Array{...}}`). The solvers get values and then iteratively `push!`. So `push!` is crucial to stay at the same speed.

But then next is indexing. The solvers and event handlers only actually use the linear index, and that should stay the same speed because that's just passing through to indexing the vector.

The more general indexing is only used on the user side. It's a convenience, and so is the conversion to a matrix function (maybe we should have it just be `full` instead of this new function?). It would be good to know what exactly the difference is when doing operations like this, but since it's not in the main routine we have more leeway here.

ChrisRackauckas:
I took a good look at this. I'm not sure that the extra fields are really worthwhile. The `ragged` field is only used once:

```julia 
S.ragged && throw(BoundsError("A ragged VectorOfArray does not support Cartesian indexing"))
```

But that's unnecessary: in many cases a Cartesian index still makes sense. For example, with this you can't do `a[1,:]` to get the first component through time, which will make sense for any (standard indices) ragged array. It think it would be better to just let the data throw the index out of bounds if the user hits something that it can't handle. Because of this, I think ragged checks and saving of `S.ragged` can all go, since that's the only place it's actually used.

But that also means you can just take `dims` as `size(S) == size(S.data)`, since that was just to support the `ragged` stuff. So I think you only need `S.data`.

And then once you do this, the overhead it's a zero-overhead abstraction too. It would still be interesting to test the cost in indexing, but that would be a structural cost and not something you could improve at all.

Even after you trim all of this back, I don't think you hit any of the functionality in the tests, which shows that it wasn't essential anyways.

I rebased you to the current RecursiveArrayTools.jl since a bit has changed here.

gabrielgellner:
Okay this Friday I will look over this. I really hope that removing ragged and dims makes everything awesome, which makes sense! I never liked having the ragged check inside what would be a very commonly used indexing operation (at the user side).

gabrielgellner:
When you did your test did you remove the subtyping from AbstractArray? I am getting a bunch of issues from the dims not really being what we would want, since `size(a.data)` gives the container vector dimension, but is not really what we would want for something like `[[1, 2], [3, 4]]` which should be like a matrix.

ChrisRackauckas:
>When you did your test did you remove the subtyping from AbstractArray? I am getting a bunch of issues from the dims not really being what we would want, since size(a.data) gives the container vector dimension, but is not really what we would want for something like [[1, 2], [3, 4]] which should be like a matrix.

Why not just define dims? When it's subtyped as `AbstractArray{T,N}`, is that actually inferable? I would think the computation on the `N` would make it type unstable to create these when it's subtyped as AbstractArray, and since you're overloading the indexing functions anyways, it may not make sense to subtype that.

ChrisRackauckas:
@gabrielgellner will you be available tomorrow? I would like to work through this and finish this up over the weekend.

gabrielgellner:
Saturday is my wife's birthday. But I plan to power through this on Sunday if that works.

ChrisRackauckas:
Sounds good.

src/vector_of_array.jl
```diff
@@ -1,56 +1,47 @@
...
-Base.size(S::VectorOfArray) = (size(S.data[1])..., length(S.data))
+Base.endof(VA::VectorOfArray) = endof(VA.data)
+Base.size(VA::VectorOfArray) = (size(VA.data[1])..., length(VA.data))
+#TODO: should we redefine length to be over the VA.data? Currently it is the number of total elements
 
-@inline function Base.getindex{T, N}(S::VectorOfArray{T, N}, I::Vararg{Int, N})
-    @boundscheck checkbounds(S, I...) # is this needed?
-    S.data[I[end]][Base.front(I)...]
+@inline function Base.getindex{T, N}(VA::VectorOfArray{T, N}, I::Vararg{Int, N})
+    @boundscheck checkbounds(VA, I...) # is this needed?
```
ChrisRackauckas: I don't think this is needed

src/vector_of_array.jl
```diff
@@ -1,56 +1,47 @@
...
-  #TODO: I don't really use dims, how do I get rid of it, and still have everything work?
-  dims::NTuple{N, Int}
 end
 
-function VectorOfArray(vec::AbstractVector)
-    # Only allow a vector of equally shaped subtypes, this will not work for ragged arrays
-    #@assert all(size(vec[1]) == size(v) for v in vec)
-    VectorOfArray(vec, (size(vec[1])..., length(vec)))
-end
+VectorOfArray{T, N}(vec::AbstractVector{T}, dims::NTuple{N}) = VectorOfArray{eltype(T), N, typeof(vec)}(vec)
```
ChrisRackauckas: Is this pure? Can this be inferred?

ChrisRackauckas:
```julia
type VectorOfArray{T, N, A} <: AbstractArray{T, N}
    data::A
end
VectorOfArray{T, N}(vec::AbstractVector{T}, dims::NTuple{N}) = VectorOfArray{eltype(T), N, typeof(vec)}(vec)
VectorOfArray(vec::AbstractVector) = VectorOfArray(vec, (size(vec[1])..., length(vec)))

A = rand(4)
@code_warntype VectorOfArray(A)


Variables:
  #self#::Type{VectorOfArray}
  vec::Array{Float64,1}

Body:
  begin 
      (Base.arrayref)(vec::Array{Float64,1},1)::Float64
      return $(Expr(:new, VectorOfArray{Float64,1,Array{Float64,1}}, :(vec)))
  end::VectorOfArray{Float64,1,Array{Float64,1}}
```

Great! That infers well. I think you cracked the code for how to make it type-inferably `<:AbstractArray`!

ChrisRackauckas:
```
TODO: should we redefine length to be over the VA.data? Currently it is the number of total elements
```

What do you think? From a quick search over DiffEq, it seems that re-defining length wouldn't be too disruptive (it was defined before on `sol` as the length of the underlying vector). What do you think is more useful? I think it might help if you're iterating over the linear index, because then the indices are `1:length(sol)`.

Other thing: this should all be written on some `AbstractVectorOfArray` instead of on the concrete type. We should probably standardize this all to use `.u`. Then the `DiffEqVectorOfArray <: AbstractVectorOfArray` with `.u` and `.t` would immediately work. The next question would be: should `DESolution <: AbstractVectorOfArray` as well? Or should it just pass indices?

gabrielgellner:
My gut feeling is that it should be over the container vector, as like you suggest I think of the `length` part as being associated with the linear index, whereas the size is over the Cartesian view. In many ways I see what we are doing as having two parts:

1. We are making a nice growable array, like you did originally that has a convenient getindex overload
2. We are making iteration, linear indexing, etc as being over the collection of subarrays 

I really think these two things together are super convenient and nice.

Also happy to make this all an Abstract type, though I don't have a great feeling over whether the diffeq solution should be a subtype. I guess it is likely a VectorOfArray with extra fields so it does make conceptual sense to me.

gabrielgellner:
One other thing I wanted feedback on. So when the array is not ragged we get sweet printing, just like it was a dense array. Once it is ragged the printing becomes meaningless. As a result I was planning on making our own print, that just gives that standard array of array view. Make sense?

ChrisRackauckas:
That makes sense to me.

ChrisRackauckas:
So just the show methods now?

gabrielgellner:
Yup. Working on that now, and then should be ready to merge.

ChrisRackauckas:
I think the show methods can come in another PR. I like short PRs better anyways. I changed `data` to `u` and will merge when the tests are complete, then test what happens if I subtype `DESolution` as an `AbstractVectorOfArray`. I think that will work just fine, and thus will expand that whole solution interface. I'm not sure `show` should be overrided on the abstract type because of this, so I'd just put it on the concrete type.

gabrielgellner:
Nice. I will keep looking into how the display machinery works, and get this ready for a future PR. Once this is merged what do you need from me next? Should I start working on adding an interp output PR?

ChrisRackauckas:
>Should I start working on adding an interp output PR?

I think that's the way to go next.

PR #6, opened 2017-05-09
Fix deprecation warnings on v0.6

ScottPJones:


src/array_partition.jl
```diff
@@ -24,30 +24,34 @@ function Base.ones(A::ArrayPartition)
...
-Base.:+(A::Number, B::ArrayPartition) = ArrayPartition((A .+ x for x in B.x)...)
-Base.:+(A::ArrayPartition, B::Number) = ArrayPartition((B .+ x for x in A.x)...)
-Base.:-(A::ArrayPartition, B::ArrayPartition) = ArrayPartition((x .- y for (x,y) in zip(A.x,B.x))...)
-Base.:-(A::Number, B::ArrayPartition) = ArrayPartition((A .- x for x in B.x)...)
-Base.:-(A::ArrayPartition, B::Number) = ArrayPartition((x .- B for x in A.x)...)
-Base.:*(A::Number, B::ArrayPartition) = ArrayPartition((A .* x for x in B.x)...)
-Base.:*(A::ArrayPartition, B::Number) = ArrayPartition((x .* B for x in A.x)...)
-Base.:/(A::ArrayPartition, B::Number) = ArrayPartition((x ./ B for x in A.x)...)
-Base.:\(A::Number, B::ArrayPartition) = ArrayPartition((x ./ A for x in B.x)...)
+import Base: +, -, *, /, \
```
ChrisRackauckas: No, I prefer the other way these days. I seems less prone to shadowing errors.

ChrisRackauckas:
Can you re-enable nightlies on Travis / AppVeyor (keep it allowed failure though). I'd like to see this does get rid of the depwarns.

src/array_partition.jl
```diff
@@ -24,31 +24,33 @@ function Base.ones(A::ArrayPartition)
...
-  Base.:.-(A::Number, B::ArrayPartition) = ArrayPartition((A .- x for x in B.x)...)
-  Base.:.-(A::ArrayPartition, B::Number) = ArrayPartition((x .- B for x in A.x)...)
-  Base.:.*(A::ArrayPartition, B::ArrayPartition) = ArrayPartition((x .* y for (x,y) in zip(A.x,B.x))...)
-  Base.:.*(A::Number, B::ArrayPartition) = ArrayPartition((A .* x for x in B.x)...)
-  Base.:.*(A::ArrayPartition, B::Number) = ArrayPartition((x .* B for x in A.x)...)
-  Base.:./(A::ArrayPartition, B::ArrayPartition) = ArrayPartition((x ./ y for (x,y) in zip(A.x,B.x))...)
-  Base.:./(A::ArrayPartition, B::Number) = ArrayPartition((x ./ B for x in A.x)...)
-  Base.:.\(A::Number, B::ArrayPartition) = ArrayPartition((x ./ A for x in B.x)...)
-end
+VERSION < v"0.6.0-dev.1614" && include_string("""
```
rdeits: I think this can be done with `@static if` rather than `include_string()`, right?

ScottPJones: No, even with `@static` it has to be parsed, and these warnings are coming directly from the Scheme parsing code, not from deprecations.jl.

rdeits: Yeah, that's what I thought too. Have you actually tried it? Adding `@static` consistently stops the deprecation warnings in v0.6 for me.

rdeits: Err, sorry, didn't notice that you said you tried it, sorry!

Then I'm super confused, because I just tried it now and it worked.

rdeits: Specifically, I changed the code to:

```julia
@static if VERSION < v"0.6-"
  Base.:.+(A::ArrayPartition, B::ArrayPartition) = ArrayPartition((x .+ y for (x,y) in zip(A.x,B.x))...)
  <other methods omitted>
end
```

and got no deprecation warnings after changing the code and re-importing it in a new Julia session.

rdeits: Or, even more simply, from the command line:

```sh
$ julia-0.6 -e "VERSION < v\"0.6-\" ? Base.:.+(x, y) = 5 : nothing"

WARNING: deprecated syntax "function Base..+(...)".
Use "function Base.broadcast(::typeof(Base.+), ...)" instead.

$ julia-0.6 -e "@static VERSION < v\"0.6-\" ? Base.:.+(x, y) = 5 : nothing"
$ 
```

ChrisRackauckas: Wait, so is the include_string necessary? I'd like to hear back from @ScottPJones first before merging. I would've assumed it wasn't necessary, and can't find a simple test where it is.

rdeits: as far as I can tell, it's not necessary, but yeah, I'm curious why @ScottPJones and I are getting different results

ScottPJones:
(everything was passing, just rebased to pick up rdeits changes for my testing)

src/array_partition.jl
ScottPJones: It didn't work for me (on Travis) without using include_string.  (I wouldn't have gone to such a convoluted method if it had worked on Travis).  In the past, I'd had to use the technique of including a small file that had only the v0.5 or v0.6 specific code, Tim Holy suggested using `include_string` here.

ScottPJones: Note: it had seemed to work for me also, just using `@static`, but then I still got the warnings on Travis.

ChrisRackauckas: Yeah, I see that

https://travis-ci.org/JuliaDiffEq/RecursiveArrayTools.jl/jobs/230412103

Not sure why. Odd that it worked locally but not on Travis. While this is slightly more convoluted, it's harmless.

PR #9, opened 2017-05-09
use @static to silence warnings in v0.6

rdeits:
Without `@static` these definitions still result in deprecation warnings on v0.6.


ChrisRackauckas:
See https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/pull/6

rdeits:
ah, thanks. I left a comment over there.

PR #18, opened 2017-08-15
Improved ArrayPartition

devmotion:
When I tried to use `ArrayPartitions` I discovered https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/11 and the closed issue https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/8 about type instabilities. However, there are still many type instabilities left - e.g. results of `zeros(x)`, `last(x)`, and all vector space operations such as multiplication and addition can not be inferred. Moreover, not in-place broadcasting does only work if the types are not changed by the applied function, as noted in https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/11.

Hence I tried to improve `ArrayPartition` and fix most of these things. This PR should fix https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/11 (all broadcasts should work in a type-stable fashion) and https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/issues/17 (`ArrayPartition` is now a subtype of `AbstractVector{T}`). Often types could not be inferred since generators were used, hence I replaced these implementations with a call to a generated function that acts as a simple unrolled generator of `ArrayPartition`. Furthermore, I tried to get rid of most type instabilities. Since I could not achieve a type-stable version of e.g. `similar(x, (Int64, Float64))` I changed this non-standard method of `similar` to `similar(x, Int64, Float64)` since then it can be made type-stable (the problem is
```julia
julia> typeof((Int64,Float64))
Tuple{DataType,DataType}
```
i.e. the type signature of the definition with a tuple contains less information).

I was wondering about the (unchanged) behaviour of `ones(x)`: in particular it also works for units (which is good, I guess), although `ones` usually does not work for arrays of units - this leads to
```julia
julia> ones([1u"m", 2u"m"])
ERROR: DimensionError: m and 1 are not dimensionally compatible.
Stacktrace:
 [1] fill!(::Array{Quantity{Int64, Dimensions:{𝐋}, Units:{m}},1}, ::Int64) at ./multidimensional.jl:788
 [2] ones(::Array{Quantity{Int64, Dimensions:{𝐋}, Units:{m}},1}, ::Type{T} where T) at ./array.jl:261 (repeats 2 times)
julia> ones(ArrayPartition([1u"m", 2u"m"]))
RecursiveArrayTools.ArrayPartition{Quantity{Int64, Dimensions:{𝐋:{𝐋
2-element Array{Quantity{Int64, Dimensions:{𝐋
 1 m
 1 m
```
which seems a bit surprising. Moreover, I do not know whether it is reasonable to define `recursive_one`  for `ArrayPartition`s by only considering the first partition. Maybe it should not be defined?

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -195,6 +195,10 @@ Base.done(A::ArrayPartition,state) = done(Chain(A.x),state)
 Base.length(A::ArrayPartition) = sum((length(x) for x in A.x))
 Base.size(A::ArrayPartition) = (length(A),)
 
+# redefine first and last to avoid slow and not type-stable indexing
+Base.first(A::ArrayPartition) = first(first(A.x))
+Base.last(A::ArrayPartition) = last(last(A.x))
+
 ## display
 
 # restore the type rendering in Juno
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -182,9 +182,11 @@ function recursivecopy!(A::ArrayPartition, B::ArrayPartition)
   end
 end
 
-recursive_one(A::ArrayPartition) = recursive_one(first(A.x))
-
 recursive_mean(A::ArrayPartition) = mean((recursive_mean(x) for x in A.x))
+
+# note: consider only first partition for recursive one and eltype
+recursive_one(A::ArrayPartition) = recursive_one(first(A.x))
+recursive_eltype(A::ArrayPartition) = recursive_eltype(first(A.x))
 
 ## iteration
 
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -1,51 +1,129 @@
-immutable ArrayPartition{T} <: AbstractVector{Any}
-  x::T
+struct ArrayPartition{T,S<:Tuple} <: AbstractVector{T}
+  x::S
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
One thing to note is that `ArrayPartition`s should be general enough to to handle heterogeneous arrays:

https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/147

Heterogeneity due to units in second order ODEs is one common area where this is used and I'd like it to still work. The goal of course is the indexing and everything is type stable when there's no units, but in the units case it's fine if just broadcasting is type stable. I'll be keeping this in mind as I review.

ChrisRackauckas:
Wow, this looks fantastic. I'm surprised you got everything to infer so well. Impressive. Did you test it a second order ODE with units?

devmotion:
No, so far I just did inference tests on `ArrayPartitions` (as shown in the tests) and compared some timings and allocations with master. For most of the operations, benchmarks were almost equal to master (often one allocation less) but the added `last` and colon operations `A[:]` are improved.

Issue #19, opened 2017-08-24
Immutable FieldVectors from StaticArrays.jl


ranocha:
`SArray` from `StaticArrays` works fine, but using a custom immutable type which is a subtype of `FieldVector`, `recursivecopy!` fails.

```
using StaticArrays
using RecursiveArrayTools

a = [SVector(1.,2.)]
b = similar(a)
recursivecopy!(b, a)


struct Bad <: FieldVector{2,Float64}
    a::Float64
    b::Float64
end

a = [Bad(1.,2.)]
b = similar(a)
recursivecopy!(b, a)
```
This results in
```
type Bad is immutable

Stacktrace:
 [1] copy!(::Bad, ::Bad) at ./multidimensional.jl:805
 [2] recursivecopy!(::Array{Bad,1}, ::Array{Bad,1}) at /.../.julia/v0.6/RecursiveArrayTools/src/utils.jl:21
 [3] include_string(::String, ::String) at ./loading.jl:515
```
According to the documentation of `StaticArrays.jl`
> it is also worth noting that `FieldVector`s may be mutable or immutable

Thus, the following works fine.
```
mutable struct NotBad <: FieldVector{2,Float64}
    a::Float64
    b::Float64
end
a = [NotBad(1.,2.)]
b = zeros(a)
recursivecopy!(b, a)
```

I came across this while I was using `OrdinaryDiffEq.jl` with `u` of the form `[Bad(1.,2.)]`. However, this has been working some months ago, see JuliaDiffEq/OrdinaryDiffEq.jl#50.

ChrisRackauckas:
I see, this needs to go down the path for `Number`.

ChrisRackauckas:
Is it not taking this path?

https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/blob/master/src/utils.jl#L9

ranocha:
No, `@which`tells me that it takes the path https://github.com/JuliaDiffEq/RecursiveArrayTools.jl/blob/master/src/utils.jl#L19

ChrisRackauckas:
I just changed master to use `StaticArray` instead of `SArray`. Trying it now.

ChrisRackauckas:
Yeah, that works, but I think then it's broken for mutable FieldVectors. I am not sure how to query for mutability.

ranocha:
Yeah, it works fine for me on `master`. Thank you!

ChrisRackauckas:
[...]

PR #21, opened 2017-10-11
fix copyat_or_push! for StaticArrays

ranocha:
This fixes some problems for OrdinaryDiffEq with callbacks if `u <: StaticArray`.


ChrisRackauckas:
Can this be specialized to SArray? This is touching on the mutability issue but I think it's important for mutable field vectors to be fast here.

ranocha:
I've implemented some additional checks as discussed in #19.

Issue #23, opened 2017-11-25
similar(v) broken for v::VectorOfArray


Gregstrq:
`similar(v)` for `v::VectorOfArray{T, N, A}` gives you an instance of `AbstractArray{T,N}` instead of another `VectorOfArray`.
MWE:
```
a = VectorOfArray([[1,2,3,4,5],[6,7,8,9,10]])
similar(a)
```
gives you a `5×2 Array{Int64,2}`.

I guess you can write something like
```
@inline Base.similar(VA::AbstractVectorOfArray) = VectorOfArray([similar(VA[i]) for i in eachindex(VA)])
```
to deal with this problem.

ChrisRackauckas:
I actually thought we already had this implemented. Yes, that is how I'd do it (along with one for DiffEq array that just passes along the `.t`). Would you mind putting in the PR?

Gregstrq:
No problem.
I am just a bit busy nowadays, so it will take several days.

Gregstrq:
It seems that VectorOfArray breaks ODE_DEFAULT_NORM function from DiffEqBase.jl/src/common_defaults (it dispatches on AbstractArray and tries to apply UNITLESS_ABS2 to subarrays of VectorOfArray).
And we can also have ArrayPartition and VectorOfArray as partition of ArrayPartition.

Seems logical to treat this cases in the same manner as for `AbstractArray{T,N} where {T<:AbstractArray,N}`. Both ArrayPartition and VectorOfArray are composite arrays, essentially.

An easy workaround would be to write something like
```
@inline DiffEqBase.ODE_DEFAULT_NORM(u::RecursiveArrayTools.ArrayPartition) = sqrt(sum(DiffEqBase.ODE_DEFAULT_NORM, u.x)/length(u.x))
@inline DiffEqBase.ODE_DEFAULT_NORM(u::RecursiveArrayTools.AbstractVectorOfArray) = sqrt(sum(DiffEqBase.ODE_DEFAULT_NORM, u)/length(u))
```
But there is still a problem when u is, for example, ArrayPartition(Array, VectorOfArray).
In this case we have an array from the first partition and arrays from VectorOfArray. So if we use the piece of code which I wrote, we will first average over each of the arrays, then we will average over averages for arrays of VectorOfArray; finally, we average over partitions of ArrayPartition.
Long story short, we don't treat arrays coming VectorOfArray and arrays coming from other partitions of ArrayPartition equally.

[...]

PR #24, opened 2017-11-28
Base.similar and Base.fill! for VectorOfArray

Gregstrq:
This PR is addressing #23.
I've added dispatch on VectorOfArray for Base.similar

I didn't do it for DiffEqArray. As I understand, DiffEqArray is used to store the results, so do we really need to use similar with it?

I've added dispatch on AbstractVectorOfArray for Base.fill!
For DiffEqArray it ignores ts and just fills u.
Now we can use zeros and ones with VectorOfArray.

I also fixed any(f, itr) for AbstractVectorOfArray. Now ODE_DEFAULT_UNSTABLE_CHECK from DiffEqBase works properly with AbstractVectorOfArray.
I suppose I should have done separate PR for it, but the change is small.

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -61,6 +61,22 @@ function Base.append!{T, N}(VA::AbstractVectorOfArray{T, N}, new_item::AbstractV
     return VA
 end
 
+# Tools for creating similar objects
+@inline Base.similar(VA::VectorOfArray{T}) where {T} = similar(VA, T)
+@inline Base.similar(VA::VectorOfArray, ::Type{T}) where {T} = VectorOfArray([similar(VA[i], T) for i in eachindex(VA)])
+
+# fill!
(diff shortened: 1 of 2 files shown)


src/vector_of_array.jl
```diff
@@ -61,6 +61,22 @@ function Base.append!{T, N}(VA::AbstractVectorOfArray{T, N}, new_item::AbstractV
     return VA
 end
 
+# Tools for creating similar objects
+@inline Base.similar(VA::VectorOfArray{T}) where {T} = similar(VA, T)
+@inline Base.similar(VA::VectorOfArray, ::Type{T}) where {T} = VectorOfArray([similar(VA[i], T) for i in eachindex(VA)])
```
ChrisRackauckas: You can put these two together with `Base.similar(VA::VectorOfArray, ::Type{T} = eltype(VA))

ChrisRackauckas:
Thanks!

PR #41, opened 2018-09-21
[WIP] Rework broadcast

vchuravy:
I think we can do away with must of the generated functions,
but as they say the proof is in pudding.

Chris, is there a set of benchmarks I should test this against?

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -3,7 +3,7 @@ struct ArrayPartition{T,S<:Tuple} <: AbstractVector{T}
 end
 
 ## constructors
-
+@inline ArrayPartition(f::F, N) where F<:Function = ArrayPartition(ntuple(f, Val(N)))
 ArrayPartition(x...) = ArrayPartition((x...,))
 
 function ArrayPartition(x::S, ::Type{Val{copy_x}}=Val{false}) where {S<:Tuple,copy_x}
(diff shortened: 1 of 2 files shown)


src/array_partition.jl
```diff
@@ -72,50 +70,32 @@ Base.ones(A::ArrayPartition, dims::NTuple{N,Int}) where {N} = ones(A)
 
 for op in (:+, :-)
```
ChrisRackauckas: Base got rid of all of these: do you think it is better to just delete these methods?

ChrisRackauckas:
Took awhile to understand but it looks good. I'm going to have to use this other places :). What's the reason for this being labelled WIP

ChrisRackauckas:
For benchmarks it would be good to retest https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/122#issuecomment-333153068

src/array_partition.jl
vchuravy: Yes absolutely.

vchuravy:
> What's the reason for this being labelled WIP?

I was waiting on you to review and I would have this see some more testing. Developing this was helpful since it pointed out issues in DistributedArrays implementation that I am working on fixing, and the example we discussed doesn't work at all for DArrays yet.
So I want to make sure that works before we merge this.

ChrisRackauckas:
Update of JuliaDiffEq/OrdinaryDiffEq.jl#122 :

```julia
using DiffEqBase, OrdinaryDiffEq, RecursiveArrayTools, DiffEqDevTools
using BenchmarkTools

u0 = zeros(100)
v0 = ones(100)
f1 = function (du,v,u,p,t)
  du .= v
end
f2 = function (dv,v,u,p,t)
  dv .= -u
end
function (::typeof(f2))(::Type{Val{:analytic}}, x, y0)
  u0, v0 = y0
  ArrayPartition(u0*cos(x) + v0*sin(x), -u0*sin(x) + v0*cos(x))
end

prob = DynamicalODEProblem(f1,f2,u0,v0,(0.0,5.0))

@benchmark solve(prob,RK4(),dt=1/2,adaptive=false,save_everystep=false)

function f(du,u,p,t)
    uu = @view u[1:100]
    v = @view u[101:end]
    du[1:100] .= uu
    du[101:200] .= v
end
u02 = [zeros(100);ones(100)]
tspan=(0.0,5.0)
prob = ODEProblem(f,u02,tspan)
@benchmark sol = solve(prob,RK4(),dt=1/2,save_everystep=false)
```

```
BenchmarkTools.Trial: 
  memory estimate:  70.98 KiB
  allocs estimate:  505
  --------------
  minimum time:     47.131 μs (0.00% GC)
  median time:      53.279 μs (0.00% GC)
  mean time:        69.464 μs (8.20% GC)
  maximum time:     2.615 ms (93.94% GC)
  --------------
  samples:          10000
  evals/sample:     1

BenchmarkTools.Trial: 
  memory estimate:  149.80 KiB
  allocs estimate:  513
  --------------
  minimum time:     67.330 μs (0.00% GC)
  median time:      75.821 μs (0.00% GC)
  mean time:        97.848 μs (10.04% GC)
  maximum time:     2.486 ms (94.94% GC)
  --------------
  samples:          10000
  evals/sample:     1
```

Wow. It's looking pretty fantastic now performance-wise especially considering the fact that there's still a broadcast performance bug for Array{Float64} which might be involved. 

Tests on OrdinaryDiffEq.jl seem to pass. I say seem to because there's something that I think is unrelated that I'm tracking down.

PR #48, opened 2019-01-20
No deep copy for numbers

devmotion:
As discussed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/95#discussion_r249277410. Tests with Measurements.jl in PuMaS still succeed. Is there anything else that should be tested prior to merging?


ChrisRackauckas:
that should be sufficient.

PR #55, opened 2019-03-17
Type agnostic, finite recursive_bottom_eltype function

Jan-LucaKlees:
Before the `recursive_bottom_eltype` function was written with only
numeric types in mind. For those, it worked. But for non-numeric types
it recused infinitely, causing a `StackOverflowError`.
Now it works on any type and is sure to recurse only a finite number of
time, thus always end and return a result.
The recursion ends, when the given argument `a` is the same as `eltype(a)`. So, as there is a limited amount of types in Julia, we will only recurse a limited amount of times.

I also added tests that cover this specific function, testing numbers
with and without units as well as chars and strings.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -76,8 +76,7 @@ end
 recursive_one(a) = recursive_one(a[1])
 recursive_one(a::T) where {T<:Number} = one(a)
 
-recursive_bottom_eltype(a) = recursive_bottom_eltype(eltype(a))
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Wow, somehow I completely didn't see this PR. Sorry!

Jan-LucaKlees:
> Wow, somehow I completely didn't see this PR. Sorry!

No worries :)

PR #56, opened 2019-03-17
Actaully test `recursive_unitless_eltype`

Jan-LucaKlees:
While working on #55 I noticed, that the tests for the
`recursive_unitless_eltype` function are not actually run. They are
written down, but without a `@test` macro their result are never
checked. So I added the macro.

### DIFF
--- a/test/utils_test.jl
+++ b/test/utils_test.jl
@@ -12,21 +12,19 @@ A = [[1 2; 3 4],[1 3;4 6],[5 6;7 8]]
            4.6666666666 6.0]
 
 A = zeros(5,5)
-recursive_unitless_eltype(A) == Float64
+@test recursive_unitless_eltype(A) == Float64
 
 using Unitful
 A = zeros(5,5)*1u"kg"
-recursive_unitless_eltype(A) == Float64
+@test recursive_unitless_eltype(A) == Float64
 AA = [zeros(5,5) for i in 1:5]
-recursive_unitless_eltype(AA) == Array{Float64,2}
+@test recursive_unitless_eltype(AA) == Array{Float64,2}
 AofA = [copy(A) for i in 1:5]
-recursive_unitless_eltype(AofA) == Array{Float64,2}
+@test recursive_unitless_eltype(AofA) == Array{Float64,2}
 AofSA = [@SVector [2.0,3.0] for i in 1:5]
-recursive_unitless_eltype(AofSA) == SVector{2,Float64}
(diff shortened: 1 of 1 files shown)


test/utils_test.jl
```diff
@@ -12,21 +12,19 @@ A = [[1 2; 3 4],[1 3;4 6],[5 6;7 8]]
...
-recursive_unitless_eltype(AofA) == Array{Float64,2}
+@test recursive_unitless_eltype(AofA) == Array{Float64,2}
 AofSA = [@SVector [2.0,3.0] for i in 1:5]
-recursive_unitless_eltype(AofSA) == SVector{2,Float64}
+@test recursive_unitless_eltype(AofSA) == SVector{2,Float64}
 AofuSA = [@SVector [2.0u"kg",3.0u"kg"] for i in 1:5]
-recursive_unitless_eltype(AofuSA) == SVector{2,Float64}
-
-@inferred recursive_unitless_eltype(AofuSA)
+@test recursive_unitless_eltype(AofuSA) == SVector{2,Float64}
```
YingboMa: `@test @inferred(recursive_unitless_eltype(AofuSA)) == SVector{2,Float64}`

Jan-LucaKlees: Hi @YingboMa, I appreciate your comment, but could you please elaborate?
If you mean to tell me that I would need to add the `@inferred` back, please explain to me why.
From my understanding, the `@inferred` is just another way of testing things. But as I added the `@test` macros, I think we don't need it anymore.

YingboMa: Sorry about that. I should have explained. `@inferred` is a stronger test than `@test recursive_unitless_eltype(AofuSA) == SVector{2,Float64}`. Consider the following case

```julia
julia> using Test

julia> foo() = rand() > 2 ? Int : Float64
foo (generic function with 1 method)

julia> @test foo() == Float64
Test Passed

julia> @inferred foo()
ERROR: return type Type{Float64} does not match inferred return type Union{Type{Float64}, Type{Int64}}
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] top-level scope at none:0
```

Jan-LucaKlees: Hi @YingboMa, thank you for the clarification! Always nice to learn :) As I understand it now, `@inferred` is used to ensure type stability.
Would it make sense to add `@inferred` to all tests?

ChrisRackauckas: as many as possible, yes it would be good to have on any test that's checking output types.

Issue #68, opened 2019-08-05
ArrayPartition copyto! dimension mismatch with Array{T,N} where N>1


nantonel:
Using Julia 1.1.1:
```julia
julia> using RecursiveArrayTools

julia> x = ArrayPartition(rand(10,10),rand(10));

julia> y = zero(x);

julia> copyto!(y,x)
ERROR: DimensionMismatch("array could not be broadcast to match destination")
Stacktrace:
 [1] copyto!(::ArrayPartition{Float64,Tuple{Array{Float64,2},Array{Float64,1}}}, ::ArrayPartition{Float64,Tuple{Array{Float64,2},Array{Float64,1}}}) at ./broadcast.jl:456
 [2] top-level scope at none:0
```

I've notice there's no error when ArrayPartition is constructed using only vectors:
```julia
julia> x = ArrayPartition(rand(2), rand(10)+im.*randn(10));

julia> y = zero(x);

julia> copyto!(y,x)
```

Seems to be related to #67, but I've opened a new issue because the error I get is different.

ChrisRackauckas:
yeah I think this is erroring for a different reason too. Not exactly trivial to fix.

lostella:
What does `zero(x)` produce? Another `ArrayPartition` with the same internal layout? In that case, a (local, temporary) fix for this could be that of implementing `copyto!` by looping through the partition pieces

ChrisRackauckas:
Actually, this just looks like a missing method.

```julia
julia> using RecursiveArrayTools

julia> x = ArrayPartition(rand(10,10),rand(10));

julia> y = zero(x)
([0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0; … ; 0.0 0.0 … 0.0 0.0; 0.0 0.0 … 0.0 0.0], [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0])

julia> typeof(zero(x))
ArrayPartition{Float64,Tuple{Array{Float64,2},Array{Float64,1}}}

julia> y .= x
[...]

PR #69, opened 2019-08-06
Fix to copyto!(::ArrayPartition,::ArrayPartition) and copyto!(::Array,::ArrayPartition)

nantonel:
Addressing #68 

I've also added a fix to the method `copyto!(::Array,::ArrayPartition)` and added some tests.

Concerning the method `copyto!(::ArrayPartition,::ArrayPartition)` I finally went for a "hybrid solution". 

Since `x.=y` is way faster that the implementation I've proposed in #68 and most of the time it is expected for `ArrayPartition`s to be partitioned equally I'm calling the `x.=y` when this is the case.

It is not so elegant but still it allows for different partitioning and adds only little overhead when `ArrayPartition`s have the same partitioning. 

```julia
using RecursiveArrayTools, BenchmarkTools
x = ArrayPartition(randn(1000,10),randn(1000),randn(3)); 
y =zero(x); 
@btime copyto!($y,$x);
@btime $y .= $x;
  2.954 μs (11 allocations: 192 bytes)
  2.923 μs (9 allocations: 160 bytes)
```

Let me know if you think there is room for improvement!


test.jl
```diff
@@ -0,0 +1,23 @@
+using RecursiveArrayTools, BenchmarkTools
+
+function Base.copyto!(A::ArrayPartition,src::ArrayPartition)
```
ChrisRackauckas: what's this file for?

src/array_partition.jl
```diff
@@ -107,18 +107,25 @@ function Base.copyto!(dest::Array,A::ArrayPartition)
...
     dest
 end
 
 function Base.copyto!(A::ArrayPartition,src::ArrayPartition)
     @assert length(src) == length(A)
-    cur = 1
-    @inbounds for i in 1:length(A.x)
-        A.x[i] .= @view(src[cur:(cur+length(A.x[i])-1)])
-        cur += length(A.x[i])
+    if size.(A.x) == size.(src.x)
```
ChrisRackauckas: does this need `if all(...)`?

test.jl
nantonel: sorry just used for benchmarking, will remove.

src/array_partition.jl
nantonel: It seems to be working without `all`... I guess when using `all` you would need `.==`?

nantonel: It seems we have to use `size.(A.x) == size.(src.x)`. If partition don't agree this can happen:
```julia
julia> x = ArrayPartition(randn(3,2),randn(2)); y = ArrayPartition(randn(3),randn(3),randn(2));

julia> all(size.(x.x) .== size.(y.x))
ERROR: DimensionMismatch("arrays could not be broadcast to a common size")
```

src/array_partition.jl
```diff
@@ -103,22 +103,29 @@ end
 Base.mapreduce(f,op,A::ArrayPartition) = mapreduce(f,op,(mapreduce(f,op,x) for x in A.x))
 Base.any(f,A::ArrayPartition) = any(f,(any(f,x) for x in A.x))
 Base.any(f::Function,A::ArrayPartition) = any(f,(any(f,x) for x in A.x))
-function Base.copyto!(dest::Array,A::ArrayPartition)
+function Base.copyto!(dest::AbstractArray,A::ArrayPartition)
```
nantonel: this seems more general

ChrisRackauckas:
Thanks!

Issue #71, opened 2019-08-27
broadcast assign an array to ArrayPartition


goretkin:
```julia
import RecursiveArrayTools: ArrayPartition

a = ArrayPartition([1], [2])

# works
a[1] = 10
a[2] = 20

# does not
a .= [10, 20]
```

with `DimensionMismatch("destination axes (Base.OneTo(1),) are not compatible with source axes (Base.OneTo(2),)")`

Issue #67, opened 2019-08-01
Error when broadcast-assigning scalar to ArrayPartition


jlumpe:
I don't know enough about broadcasting or the AbstractArray interface to know what's actually going on here. This is with version 1.0.0.

```
julia> a = ArrayPartition(zeros(10))
([0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],)

julia> a .= 1.
ERROR: MethodError: copyto!(::ArrayPartition{Float64,Tuple{Array{Float64,1}}}, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Tuple{Base.OneTo{Int64}},typeof(identity),Tuple{Float64}}) is ambiguous. Candidates:
  copyto!(dest::ArrayPartition, bc::Base.Broadcast.Broadcasted) in RecursiveArrayTools at /Users/student/.julia/packages/RecursiveArrayTools/vxlTS/src/array_partition.jl:247
  copyto!(dest::AbstractArray, bc::Base.Broadcast.Broadcasted{#s623,Axes,F,Args} where Args<:Tuple where F where Axes where #s623<:Base.Broadcast.AbstractArrayStyle{0}) in Base.Broadcast at broadcast.jl:803
Possible fix, define
  copyto!(::ArrayPartition, ::Base.Broadcast.Broadcasted{#s623,Axes,F,Args} where Args<:Tuple where F where Axes where #s623<:Base.Broadcast.AbstractArrayStyle{0})
Stacktrace:
 [1] materialize!(::ArrayPartition{Float64,Tuple{Array{Float64,1}}}, ::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{0},Nothing,typeof(identity),Tuple{Float64}}) at ./broadcast.jl:756
 [2] top-level scope at none:0
```

ChrisRackauckas:
Not many people know enough about broadcasting for this. This might be pretty hard to untangle.

Issue #60, opened 2019-06-16
Outer product of ArrayPartitions do not produce "matrix like" objects.


pkofod:
Expected: expected outer product produces something of shape `(n, n)' for two n-dimensional ArrayPartitions.
Got: `(n*n,)` shaped array of partial outer products.

```
julia> using RecursiveArrayTools

julia> xap = ArrayPartition([1.0], [2.0, 3.0])
([1.0], [2.0, 3.0])

julia> xap*xap'
([1.0 2.0 3.0], [2.0 4.0 6.0; 3.0 6.0 9.0])

julia> size(ans)
(9,)
```

Issue #44, opened 2018-10-25
Some DifferentialEquations.jl solvers fail when using `ArrayPartition`


Sleort:
Note the use of `ArrayPartition` below
```
julia> using DifferentialEquations,RecursiveArrayTools

julia> function lorenz(du,u,p,t)
           du[1] = 10.0*(u[2]-u[1])
           du[2] = u[1]*(28.0-u[3]) - u[2]
           du[3] = u[1]*u[2] - (8/3)*u[3]
       end
lorenz (generic function with 1 method)

julia> u0 = ArrayPartition([1.0,0.0],[0.0])
([1.0, 0.0], [0.0])

julia> tspan = (0.0,100.0)
(0.0, 100.0)

julia> prob = ODEProblem(lorenz,u0,tspan)
ODEProblem with uType ArrayPartition{Float64,Tuple{Array{Float64,1},Array{Float64,1}}} and tType Float64. In-place: true
timespan: (0.0, 100.0)
u0: [1.0, 0.0][0.0]

julia> sol = solve(prob,Tsit5());

julia> sol = solve(prob,AutoTsit5(Rosenbrock23()));
ERROR: MethodError: no method matching ldiv!(::LinearAlgebra.LU{Float64,Array{Float64,2}}, ::ArrayPartition{Float64,Tuple{Array{Float64,1},Array{Float64,1}}})
Closest candidates are:
  ldiv!(::LinearAlgebra.Transpose{#s561,#s560} where #s560<:LinearAlgebra.LowerTriangular where #s561, ::AbstractArray{T,1} where T) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/triangular.jl:1217
  ldiv!(::LinearAlgebra.Transpose{#s561,#s560} where #s560<:LinearAlgebra.LowerTriangular where #s561, ::AbstractArray{T,1} where T, ::AbstractArray{T,1} where T) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/triangular.jl:1201
  ldiv!(::LinearAlgebra.Transpose{#s561,#s560} where #s560<:LinearAlgebra.UnitLowerTriangular where #s561, ::AbstractArray{T,1} where T) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/LinearAlgebra/src/triangular.jl:1235
  ...
```
[...]

PR #88, opened 2020-04-07
support linear algebra on ArrayPartition

ChrisRackauckas:
Should get improved, but the general idea. The hard part for the stiff ODE solvers seems to be the broadcast

Fixes https://github.com/SciML/RecursiveArrayTools.jl/issues/71 Fixes https://github.com/SciML/RecursiveArrayTools.jl/issues/67 Fixes https://github.com/SciML/RecursiveArrayTools.jl/issues/60 Fixes https://github.com/SciML/RecursiveArrayTools.jl/issues/44

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,11 @@ language: julia
 os:
   - linux
 julia:
-  - 1.0
-  - nightly
-matrix:
-  allow_failures:
-    - julia: nightly
+  - 1
+#  - nightly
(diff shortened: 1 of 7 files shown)


ChrisRackauckas:
A broadcast overload which allows for mixing in Array is required to fix autodiff Jacobians, which can probably be done with a new interface function for `stridedvec` but can be done better through a broadcast overload.

Issue #72, opened 2019-09-16
allocating linearalgebra


rveltz:
Hi,

I am surprised by the following behavior. Can you please tell me if this is expected?

Thank you,

Best regards,

```
using BenchmarkTools

xce0 = ArrayPartition(zeros(2),[0.])
xcde0 = copy(xce0)

function foo(y, x)
	y .= y .+ 2 .* x
end

@btime foo($xcde0, $xce0)
```
which gives

```
julia> @btime foo($xcde0, $xce0)
  30.138 ns (3 allocations: 48 bytes)
```

ChrisRackauckas:
It's 48 bytes, so it could just be the fact that you're returning it. What happens if you add nothing to the end?

rveltz:
Nothing :(

What's more:

```
function foo(y, x)
	y .= y .+ 2 .* x + 3 .* x
	nothing
end

@btime foo($xcde0, $xce0)
```
gives `243.377 ns (19 allocations: 832 bytes)`

ChrisRackauckas:
Uhh, can't say I have an explanation for that. I wonder if there's something in the broadcast definition we should be avoiding. Whatever it is, this will take time.

PR #89, opened 2020-04-07
fix broadcast allocations

ChrisRackauckas:
```julia
using RecursiveArrayTools, BenchmarkTools
xce0 = ArrayPartition(zeros(2),[0.])
xcde0 = copy(xce0)
function foo(y, x)
	y .= y .+ 2 .* x
end
@btime foo($xcde0, $xce0)
```

```
19.075 ns (0 allocations: 0 bytes)
```

Fixes https://github.com/SciML/RecursiveArrayTools.jl/issues/72 

@rveltz


rveltz:
thx

PR #100, opened 2020-05-27
hotfix VectorOfArray Cartesian indexing

ChrisRackauckas:


ChrisRackauckas:
Downstream passes locally, and the extra broadcast tests cover the behavior that was found.

PR #113, opened 2020-08-31
Added linalg mul! overloads for ArrayPartition with tests

Lewih:
First commit

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -372,16 +372,27 @@ function LinearAlgebra._swap_rows!(B::ArrayPartition, i::Integer, j::Integer)
 end
 
 # linalg mul! overloads for ArrayPartition
-function LinearAlgebra.mul!(C::T, A::T, B::AbstractArray) where T<:ArrayPartition
-    @assert length(C.x) == length(A.x)
+function LinearAlgebra.mul!(C::ArrayPartition, A::ArrayPartition, B::AbstractArray)
+    if length(C.x) != length(A.x)
+        throw(DimensionMismatch("Length of C, $(length(C.x)), does not match length of A, $(length(A.x))"))
+    end
+
     for index = 1:length(C.x)
         mul!(C.x[index], A.x[index], B)
     end
+    return C
 end
 
-function LinearAlgebra.mul!(C::T, A::T, B::T) where T<:ArrayPartition
-    @assert length(C.x) == length(A.x) == length(B.x)
+function LinearAlgebra.mul!(C::ArrayPartition, A::ArrayPartition, B::ArrayPartition)
+    if length(C.x) != length(A.x)
+        throw(DimensionMismatch("Length of C, $(length(C.x)), does not match length of A, $(length(B.x))"))
+    end
+    if length(A.x) != length(B.x)
+        throw(DimensionMismatch("Length of A, $(length(A.x)), does not match length of B, $(length(B.x))"))
+    end
+
     for index = 1:length(C.x)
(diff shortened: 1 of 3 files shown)


src/array_partition.jl
```diff
@@ -370,3 +370,18 @@ function LinearAlgebra._swap_rows!(B::ArrayPartition, i::Integer, j::Integer)
     B[i], B[j] = B[j], B[i]
     return B
 end
+
+# linalg mul! overloads for ArrayPartition
+function LinearAlgebra.mul!(C::T, A::T, B::AbstractArray) where T<:ArrayPartition
+    @assert length(C.x) == length(A.x)
```
devmotion, line 376: According to the docstring, "An assert might be disabled at various optimization levels. Assert should therefore only be used as a debugging tool". Hence maybe change it to a regular check and explicitly throw an error?

src/array_partition.jl
```diff
@@ -370,3 +370,18 @@ function LinearAlgebra._swap_rows!(B::ArrayPartition, i::Integer, j::Integer)
...
+    for index = 1:length(C.x)
+        mul!(C.x[index], A.x[index], B)
+    end
+end
+
+function LinearAlgebra.mul!(C::T, A::T, B::T) where T<:ArrayPartition
+    @assert length(C.x) == length(A.x) == length(B.x)
```
devmotion, line 383: Same here?

test/linalg.jl
```diff
@@ -1,5 +1,5 @@
 using RecursiveArrayTools, Test, Random
-using LinearAlgebra
+using LinearAlgebra, SparseArrays
```
Lewih, line 3: Probably Sparse Array test is not necessary

test/linalg.jl
```diff
@@ -26,3 +26,24 @@ for ff in (lu, svd, qr)
...
+n, m, l = 5, 6, 7
+bb = rand(n, n), rand(m, n), rand(l, n)
+cc = rand(n), rand(n), rand(n)
+dd = rand(n), rand(m), rand(l)
+b = ArrayPartition(bb)
+c = ArrayPartition(cc)
+d = ArrayPartition(dd)
+A = rand(n)
+for T in (Array{Float64}, Array{ComplexF64},)
+    B = T(A)
```
devmotion, line 40: Without `local` a soft scope warning is printed since it seems a global variable `B` already exists:
```suggestion
    local B = T(A)
```
The same problem exists in line 10 BTW.

Lewih: We should fix the whole test file at this point in order to avoid this tedious warnings.
Next commit I'll have polished the file.

src/array_partition.jl
```diff
@@ -370,3 +370,29 @@ function LinearAlgebra._swap_rows!(B::ArrayPartition, i::Integer, j::Integer)
     B[i], B[j] = B[j], B[i]
     return B
 end
+
+# linalg mul! overloads for ArrayPartition
+function LinearAlgebra.mul!(C::ArrayPartition, A::ArrayPartition, B::AbstractArray)
```
ChrisRackauckas, line 375: this should probably check for size and either do views or all of `B`?

Lewih: do you mean something like:
```julia
function LinearAlgebra.mul!(C::ArrayPartition{T, S}, A::ArrayPartition{T, S}, B::AbstractArray{T, S}) where T<:Number
```

ChrisRackauckas: I was out of my mind and it didn't make too much sense.

PR #119, opened 2020-11-02
Define the custom adjoint on a more general ArrayPartition constructor

cqql:
Computing the adjoint of a `SecondOrderODEProblem` requires an adjoint for this constructor because of the following line.

https://github.com/SciML/DiffEqBase.jl/blob/9f0875ad24ac9c2f5784b242eba5344807f278e7/src/problems/ode_problems.jl#L153

This PR depends on FluxML/ZygoteRules.jl#12 because of the anonymous value type argument. That was a fun one to debug :)

### DIFF
--- a/src/zygote.jl
+++ b/src/zygote.jl
@@ -15,13 +15,14 @@ ZygoteRules.@adjoint function getindex(VA::AbstractVectorOfArray, i, j...)
   VA[i,j...],AbstractVectorOfArray_getindex_adjoint
 end
 
-ZygoteRules.@adjoint function ArrayPartition(x...)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Thanks!

ChrisRackauckas:
https://travis-ci.org/github/SciML/RecursiveArrayTools.jl/builds/740939125#L439 looks like it needs to hit the 3 argument version.

cqql:
No, this error is the reason why the other PR needs to be merged first. At the moment, the generated `ZygoteRules._pullback` passes the second argument wrongly to `ZygoteRules.adjoint` (the `Type{Type{...}}` construct in the error message).

cqql:
Now that v0.2.1 of ZygoteRules has been released, this PR is ready to be merged as well.

PR #123, opened 2020-12-04
bump compat version bounds

ranocha:
If tests pass, there should be a new release of RecursiveArrayTools.jl since it holds back updates for NonlinearSolve.jl (https://github.com/JuliaComputing/NonlinearSolve.jl/pull/16) and DiffEqBase.jl (SciML/DiffEqBase.jl#604) which in turns hold back updates for a lot of downstream packages.

This needs a new release of OrdinaryDiffEq.jl.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -16,7 +16,7 @@ ZygoteRules = "700de1a5-db45-46bc-99cf-38207098b444"
 ArrayInterface = "2.7"
 RecipesBase = "0.7, 0.8, 1.0"
 Requires = "0.5, 1.0"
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
It's failing because of how the downstream dependencies are being handled. I took a look through the release and I don't think we should have any issues: the breaking part is the depwarn removals. So I'm going to do the scary thing of merging this, tagging, and letting things update. Then let's make sure we re-run this test to see that we didn't miss anything, but the library wasn't throwing depwarns so it should be good.

PR #126, opened 2020-12-10
Fixes #125

### LINKED ISSUE #125: Vector of StaticArrays inside ArrayPartition
I have tried to use a `Vector` of `StaticArray`s inside an `ArrayPartition`. This led to a non-trivial problem with DifferentialEquations.jl, when I tried to use such an `ArrayPartition` as the state vector for an ODE: the resulting element type of `ArrayPartition` happened to be `Any` and it broke `recursive_unitless_bottom_eltype` function which is used somewhere inside DifferentialEquations.jl.

In order to illustrate this, consider the following example:
```
using RecursiveArrayTools, StaticArrays

a = ArrayPartition((zeros(Float64, 10), zeros(SVector{3,Float64}, 10)))
```
The output of `typeof(a)` is
> ArrayPartition{Any,Tuple{Array{Float64,1},Array{SArray{Tuple{3},Float64,1,3},1}}}

And if you try to do `recursive_unitless_bottom_eltype(a)`, you get a never-ending recursion:
>ERROR: StackOverflowError:
Stacktrace:
 [1] recursive_unitless_bottom_eltype(::Type{Any}) at ~/.julia/packages/RecursiveArrayTools/BVS4M/src/utils.jl:86 (repeats 79984 times)

In the example I provided, it would be logical if the element type of `ArrayPartition` is `Float64`. It seems that we can get such behaviour if we slightly change the constructor for `ArrayPartition`:
```
function ArrayPartition(x::S, ::Type{Val{copy_x}}=Val{false}) where {S<:Tuple,copy_x}
  T = promote_type(recursive_bottom_eltype.(x)...)
  if copy_x
    return ArrayPartition{T,S}(copy.(x))
  else
    return ArrayPartition{T,S}(x)
  end
end
```
The only change I propose to make is to replace `eltype` with `recursive_bottom_eltype` in the type promotion. Potentially, this solution can deal with any kind of crazy nesting of arrays.

Gregstrq:
Fixed the determination of eltype of ArrayPartition in the case of complex nested arrays, such as vectors of StaticArrays, for example.

Fixes #125

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -7,7 +7,7 @@ end
 ArrayPartition(x...) = ArrayPartition((x...,))
 
 function ArrayPartition(x::S, ::Type{Val{copy_x}}=Val{false}) where {S<:Tuple,copy_x}
-  T = promote_type(eltype.(x)...)
+  T = promote_type(recursive_bottom_eltype.(x)...)
   if copy_x
     return ArrayPartition{T,S}(copy.(x))
   else
--- a/test/partitions_and_static_arrays.jl
+++ b/test/partitions_and_static_arrays.jl
@@ -1,0 +1,10 @@
+using Test, RecursiveArrayTools, StaticArrays
+
+p = ArrayPartition((zeros(Float32, 2), zeros(SMatrix{2,2, Int64},2), zeros(SVector{3, Float64}, 2)))
+@test eltype(p)==Float64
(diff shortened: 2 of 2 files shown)


ChrisRackauckas:
Add a test?

I added the link to the issue.

Gregstrq:
I added the tests. They are passed all right.

PR #128, opened 2021-01-07
fix ArrayInterface.ismutable(::ArrayPartition)

KlausC:
A `PartitionedArray` should be considered "mutable" only, if all of its components are "mutable".
Therefore the `ArrayInterface.ismutable` has been specialized for this case.
This fixes a bug in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1322

Another question is the naming `ArrayInteface.ismutable` which collides with `Base.ismutable` and returns different results.

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -70,6 +70,11 @@ end
 # ignore dims since array partitions are vectors
 Base.ones(A::ArrayPartition, dims::NTuple{N,Int}) where {N} = ones(A)
 
+# mutable iff all components of ArrayPartition are mutable
+function ArrayInterface.ismutable(::Type{<:ArrayPartition{T,S}}) where {T,S}
+    all(ArrayInterface.ismutable, S.parameters)
+end
+
 ## vector space operations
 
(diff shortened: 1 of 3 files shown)


src/array_partition.jl
```diff
@@ -70,6 +70,11 @@ end
 # ignore dims since array partitions are vectors
 Base.ones(A::ArrayPartition, dims::NTuple{N,Int}) where {N} = ones(A)
 
+# mutable iff all components of ArrayPartition are mutable
+function ArrayInterface.ismutable(::Type{<:ArrayPartition{T,S}}) where {T,S}
```
YingboMa, line 74: This should be a generated function to make it have zero runtime.

ChrisRackauckas:
>Another question is the naming ArrayInteface.ismutable which collides with Base.ismutable and returns different results.

Yeah... ours came first and does the right thing though 😆 

Agreed with Yingbo it should be a generated function and then it's good.

ChrisRackauckas (approved):

PR #134, opened 2021-04-02
Add more specialized functions

garibarba:
- Adds an `any` method without predicate. Needed for efficiently calling `any` on e.g. `ArrayPartition{Float32, Tuple{CuArray{Float32, 1}, CuArray{Float32, 1}}}`
- Adds analogous `all` methods.


ChrisRackauckas:
Thanks!

PR #139, opened 2021-04-23
fix recursivecopy

ranocha:
As discussed in https://github.com/SciML/OrdinaryDiffEq.jl/issues/1386 and suggested by @ChrisRackauckas.

Closes https://github.com/SciML/OrdinaryDiffEq.jl/issues/1386


jlchan:
Should I add `ArrayInterface.ismutable(x::StructArray) = true` before marking https://github.com/SciML/OrdinaryDiffEq.jl/issues/1386 as closed?

ranocha:
No, you don't need that - it's the default

ranocha:
SciML/OrdinaryDiffEq.jl#1386 should be closed once this PR is merged - you don't need to do this manually, i think

jlchan:
Thanks! Much appreciated

PR #142, opened 2021-06-02
Make ZygoteRules for getindex use `zero`

lamorton:
Prelude to an actual fix for https://github.com/SciML/ModelingToolkit.jl/issues/972.  This makes the two ZygoteRules for getindex more alike, in that both now use `zero` to construct a copy of the array and then slice into it. The error message for the original problem becomes more comprehensible.


ChrisRackauckas:
Is there a way to do this that avoids mutation? The Zygote getindex really needs to be changed.

lamorton:
Yeah, I just realized this won't work anyway, when it comes to systems with observed variables.

PR #147, opened 2021-06-28
Fix for #146

### LINKED ISSUE #146: 2.13-> 2.14: broken API or breaking change ?
Between 2.13 and 2.14, the meaning of  `a[i,:,t]` changed.  I think it is broken. If it is intended to be a breaking change I would expect the version to be 3.0 ...

- [mwe](https://gist.github.com/j-fu/b2bf4ae92f5619199ca51865d12c234f#file-mwe-jl)

- output  of [2.13.0](https://gist.github.com/j-fu/b2bf4ae92f5619199ca51865d12c234f#file-out_2-13)

- output  of [2.14.0](https://gist.github.com/j-fu/b2bf4ae92f5619199ca51865d12c234f#file-out_2-14)

sharanry:
Fixes #146 

CC: @ChrisRackauckas

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -76,23 +76,6 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractDiffEqArray{T, N},sym
     Base.getindex.(A.u, i)
   end
 end
-Base.@propagate_inbounds function Base.getindex(A::AbstractDiffEqArray{T, N},sym,args...) where {T, N}
-  if issymbollike(sym) && A.syms !== nothing
-    i = findfirst(isequal(Symbol(sym)),A.syms)
-  else
-    i = sym
-  end
-
-  if i === nothing
-    if issymbollike(i) && A.indepsym !== nothing && Symbol(i) == A.indepsym
-      A.t[args...]
-    else
-      observed(A,sym,args...)
-    end
-  else
-    Base.getindex.(A.u, args...)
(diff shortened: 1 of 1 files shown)


sharanry:
This is a temporary fix that shouldn't affect anything else. Moore thorough testing will probably be required to prevent this from happening in the future.

ChrisRackauckas:
I was thinking it might just be missing an `Int...` dispatch? Something like https://github.com/SciML/RecursiveArrayTools.jl/pull/147/files#diff-a6e6da66354c0871e3f4bae3523909853bd38bdc881c3826ab93091c96eb8fd8R57 but then arbitrary layers deep.

sharanry:
I think the problem with such a dispatch is, it doesn't cover colons?

ChrisRackauckas:
Good point. It may need some of these: https://github.com/SciML/SciMLBase.jl/blob/master/src/solutions/solution_interface.jl#L20-L29

sharanry:
@ChrisRackauckas 
I am planning to add a bunch of tests based on [this](https://codecov.io/gh/SciML/RecursiveArrayTools.jl/src/master/src/vector_of_array.jl) report. The coverage has been deteriorating with time. 
Could we merge this and make these additional tests a separate PR?

ChrisRackauckas:
Yeah, let's merge and see. Today isn't going to be a clean day.

PR #148, opened 2021-06-29
Add special cases for DiffEqArray constructor

sharanry:
Required for https://github.com/SciML/SciMLBase.jl/pull/66


ChrisRackauckas:
Can we take a bit of time to expand the test suite here?

sharanry:
> Can we take a bit of time to expand the test suite here?

Yes -- was about to do it. It should make things much smoother.

PR #153, opened 2021-07-05
Broaden vector eltype signature for `push!`

stillyslalom:
Pushing to `AbstractVectorOfArray` is currently limited to vectors with eltype `AbstractVector`, which seems likely to have been a too-restrictive oversight. This PR broadens the accepted eltype to `AbstractArray`.


ChrisRackauckas (approved):
yes, nice find.

PR #156, opened 2021-07-08
Optimize issymbollike

YingboMa:
```julia
julia> @variables x;

julia> @btime issymbollike(Ref($x)[])
  0.026 ns (0 allocations: 0 bytes)
true

julia> @btime issymbollike(Ref(1)[])
  0.026 ns (0 allocations: 0 bytes)
false

julia> @btime issymbollike(Ref(:x)[])
  0.026 ns (0 allocations: 0 bytes)
true
```

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -16,20 +16,12 @@ struct AllObserved
 end
 
 Base.@pure __parameterless_type(T) = Base.typename(T).wrapper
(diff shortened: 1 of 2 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
I'm surprised `Base.@pure` didn't get to this.

PR #157, opened 2021-07-12
Remove unreachable code

sharanry:
https://github.com/SciML/SciMLBase.jl/issues/83

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -71,7 +71,7 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractDiffEqArray{T, N},sym
     i = sym
   end
 
-  if i === nothing
+  if i == nothing
     if issymbollike(sym) && A.indepsym !== nothing && Symbol(sym) == A.indepsym
       A.t
     else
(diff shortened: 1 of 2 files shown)


src/vector_of_array.jl
```diff
@@ -72,11 +72,7 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractDiffEqArray{T, N},sym
   end
 
   if i === nothing
-    if issymbollike(i) && A.indepsym !== nothing && Symbol(i) == A.indepsym
-      A.t
```
ChrisRackauckas, line 76: did you try the independent variable and see what it does?

sharanry: I think that case is covered now. Added tests too.

ChrisRackauckas (approved):

src/vector_of_array.jl
```diff
@@ -71,8 +71,8 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractDiffEqArray{T, N},sym
     i = sym
   end
 
-  if i === nothing
-    if issymbollike(i) && A.indepsym !== nothing && Symbol(i) == A.indepsym
+  if i == nothing
```
YingboMa, line 74: This is not recommended. Why do you need to use `==`?

YingboMa (commented):
Do you need `==`'s generic methods?

src/vector_of_array.jl
sharanry: Please see https://github.com/SciML/SciMLBase.jl/pull/84#discussion_r667998159 .

sharanry:
> Do you need `==`'s generic methods?

Are you referring to the same `==`/`===` issue?

PR #158, opened 2021-07-17
[WIP] remove un-roundtripable 2-arg `show` methods

anandijain:
in https://github.com/SciML/ModelingToolkit.jl/pull/1127 
I make writing Systems to file include symbolics' reserved metadata (like units, and noise).

however, the custom 2-arg pretty printing used here breaks the ability to reparse the result of `show` methods for NoiseProcesses like so `var"##sts#259"[1] = setmetadata(var"##sts#259"[1], ModelingToolkit.VariableNoiseType, t: [0] u: [0])`
where `t: [0] u: [0]` is really `WienerProcess(0, 0, 0)`

now it gets shown as `[0]` so still bad

cc @YingboMa

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -244,7 +244,6 @@ Base.last(A::ArrayPartition) = last(last(A.x))
 
 ## display
 Base.summary(A::ArrayPartition) = string(typeof(A), " with arrays:")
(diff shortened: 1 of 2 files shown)


YingboMa:
Superseded by #159

Issue #160, opened 2021-07-22
Relationship between `PartitionArray` and `ProductRepr`


mateuszbaran:
I've just noticed that `RecursiveArrayTools.jl` has something similar to `ProductRepr` from `Manifolds.jl`, and here it's called `PartitionArray`. It would be nice if both efforts could be unified to mutually gain from new developments. The main difference seems to be in the way they handle broadcasting. I don't know why they differ but the first difference I've noticed is that `ProductRepr` broadcasting is faster:
```julia
julia> a = ArrayPartition(SA[1, 2], SA[1 2; 3 4])
([1, 2], [1 2; 3 4])

julia> b = ArrayPartition(SA[1, 2], SA[1 2; 3 4])
([1, 2], [1 2; 3 4])

julia> c = ArrayPartition((@MVector [1, 2]), (@MMatrix [1 2; 3 4]))
([1, 2], [1 2; 3 4])

julia> ap = ProductRepr(SA[1, 2], SA[1 2; 3 4])
ProductRepr with 2 submanifold components:
 Component 1 =
  2-element SVector{2, Int64} with indices SOneTo(2):
   1
   2
 Component 2 =
  2×2 SMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
   1  2
   3  4

julia> bp = ProductRepr(SA[1, 2], SA[1 2; 3 4])
ProductRepr with 2 submanifold components:
 Component 1 =
  2-element SVector{2, Int64} with indices SOneTo(2):
   1
   2
 Component 2 =
  2×2 SMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
   1  2
   3  4

julia> cp = ProductRepr((@MVector [1, 2]), (@MMatrix [1 2; 3 4]))
ProductRepr with 2 submanifold components:
 Component 1 =
  2-element MVector{2, Int64} with indices SOneTo(2):
   1
   2
 Component 2 =
  2×2 MMatrix{2, 2, Int64, 4} with indices SOneTo(2)×SOneTo(2):
   1  2
   3  4

julia> f(a, b) = a .+ b
f (generic function with 1 method)

julia> f!(c, a, b) =  c.= a .+ b
f! (generic function with 1 method)

julia> @benchmark f($a, $b)
BenchmarkTools.Trial: 10000 samples with 1000 evaluations.
[...]

PR #163, opened 2021-07-23
Speed up some use cases of `ArrayPartition`

mateuszbaran:
This is the first batch of changes related to #160 . Most changes are related to the fact that `map` is much more friendly to Julia compiler than broadcasting so I've replaced a bunch of unnecessary broadcasts with `map`. The other thing is that the loop in assignment broadcast is very problematic -- ideally compiler should be able to constant-propagate `N` and unroll the loop but it seems that doing so breaks some other optimizations. This change helps a bit but it's still not enough.

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -281,7 +281,10 @@ end
 
 @inline function Base.copyto!(dest::ArrayPartition, bc::Broadcast.Broadcasted{ArrayPartitionStyle{Style}}) where Style
     N = npartitions(dest, bc)
-    ntuple(i->copyto!(dest.x[i], unpack(bc, i)), Val(N))
+    @inline function f(i)
+        copyto!(dest.x[i], unpack(bc, i))
+    end
+    ntuple(f, Val(N))
     dest
 end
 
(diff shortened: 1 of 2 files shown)


mateuszbaran:
I've improved that `copyto!` method and now my example from #160 is about 100x faster than before :tada: :
```julia
julia> @benchmark f!($c, $a, $b)
BenchmarkTools.Trial: 10000 samples with 999 evaluations.
 Range (min … max):  7.885 ns … 34.613 ns  ┊ GC (min … max): 0.00% … 0.00%
 Time  (median):     7.972 ns              ┊ GC (median):    0.00%
 Time  (mean ± σ):   7.990 ns ±  0.406 ns  ┊ GC (mean ± σ):  0.00% ± 0.00%

                                  ▂▂█▁▄▄                      
  ▂▂▂▂▁▁▂▁▂▂▁▂▂▂▂▂▂▃▃▃▃▃▃▄▄▅▄▅▄▄▇▆██████▇▇▄▅▅▄▇▆▇▄▄▃▃▂▂▃▂▃▃▃ ▃
  7.88 ns        Histogram: frequency by time         830 ns <

 Memory estimate: 0 bytes, allocs estimate: 0.
```

ChrisRackauckas (approved):
oh yes, just mapping over the tuple is a lot better, great idea.

PR #167, opened 2021-08-16
Stop recursing when the eltype is not mutable

YingboMa:


ChrisRackauckas (approved):

src/utils.jl
```diff
@@ -31,9 +31,14 @@ function recursivecopy!(b::AbstractArray{T,N},a::AbstractArray{T2,N}) where {T<:
 end
 
 function recursivecopy!(b::AbstractArray{T,N},a::AbstractArray{T2,N}) where {T<:AbstractArray,T2<:AbstractArray,N}
-  @inbounds for i in eachindex(a)
-    recursivecopy!(b[i],a[i])
+  if ArrayInterface.ismutable(T)
+    @inbounds for i in eachindex(a)
```
YingboMa, line 35:
```suggestion
    @inbounds for i in eachindex(b, a)
```

PR #168, opened 2021-09-29
Gradient definitions & supertypes

mcabbott:
This is inspired by https://github.com/SciML/DiffEqSensitivity.jl/pull/493#issuecomment-927014502 . I wrote many examples there, but the core surprise was that a `VectorOfArrays{T} <: AbstractArray{T,3}` has a gradient `Vector{Matrix{T}}` which is not a 3-array. If I understood correctly, it would desirable that the result be another VectorOfArray, i.e. something `AbstractArray{T,3}` which iterates like Vector{Matrix}.

One way to do that would be using ProjectTo. But in fact I believe the Vector{Matrix} is being explicitly generated by a rule defined here, so perhaps it is simpler just to alter that rule to wrap this?

RFC, WIP, needs tests... and needs thought about other cases:
```julia
julia> va = RecursiveArrayTools.VectorOfArray([rand(3,3), rand(3,3)]);

julia> gradient(va -> sum(va[1]), va)[1]
VectorOfArray{Float64,3}:
2-element Vector{AbstractMatrix{Float64}}:
 3×3 Fill{Float64}, with entries equal to 1.0
 [0.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]

julia> ans + va |> summary  # essential this work for a valid tangent
"VectorOfArray{Float64,3}"

julia> gradient((x,y) -> sum(x ./ y), va, rand(3))[1] |> summary  # should this also be changed?
"3×3×2 Array{Float64, 3}"

julia> gradient(va -> sum(va[:,1]), va)[1]  # this didn't work before, either. 
ERROR: BoundsError: attempt to access 3×3×2 VectorOfArray{Float64, 3, Vector{Matrix{Float64}}} at index [1:3, 1]
```
* The broadcast case could apply something like `VectorOfArray(collect(eachslice(Δ)))`, to return a comparable object, i.e. one which broadcasts one way and iterates a different way. This would need `ProjectTo`. It would also apply to things like `gradient(va -> sum(abs, [1 2; 3 4] * va), VectorOfArray([[1,2], [3,4]]))`. 
* Is indexing `va[:,1]` expected to work? What should it return? The other way around should I think do this:
```
julia> gradient(va -> sum(va[1,:]), va)[1]
VectorOfArray{Float64,3}:
2-element Vector{Matrix{Float64}}:
 [1.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]
 [1.0 0.0 0.0; 0.0 0.0 0.0; 0.0 0.0 0.0]
```

### DIFF
--- a/src/zygote.jl
+++ b/src/zygote.jl
@@ -43,10 +43,28 @@ function ChainRulesCore.rrule(::typeof(getproperty),A::ArrayPartition, s::Symbol
     A.x,literal_ArrayPartition_x_adjoint
 end
 
+#=
+
+# Define a new species of projection operator for this type:
+ChainRulesCore.ProjectTo(x::VectorOfArray) = ProjectTo{VectorOfArray}()
+
+# Gradient from iteration will be e.g. Vector{Vector}, this makes it another AbstractMatrix
+(::ProjectTo{VectorOfArray})(dx::AbstractVector{<:AbstractArray}) = VectorOfArray(dx)
+# Gradient from broadcasting will be another AbstractArray
+(::ProjectTo{VectorOfArray})(dx::AbstractArray) = dx
+
+But this may not be necessary?
+
+=#
+
+
+# These rules duplicate the `rrule` methods above, because Zygote looks for an `@adjoint`
+# definition first, and finds its own before finding those.
+
 ZygoteRules.@adjoint function getindex(VA::AbstractVectorOfArray, i::Union{Int,AbstractArray{Int},CartesianIndex,Colon,BitArray,AbstractArray{Bool}})
   function AbstractVectorOfArray_getindex_adjoint(Δ)
     Δ′ = [ (i == j ? Δ : zero(x)) for (x,j) in zip(VA.u, 1:length(VA))]
-    (Δ′,nothing)
+    (VectorOfArray(Δ′),nothing)
   end
   VA[i],AbstractVectorOfArray_getindex_adjoint
 end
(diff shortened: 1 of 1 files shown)


src/zygote.jl
```diff
@@ -55,11 +73,13 @@ ZygoteRules.@adjoint function getindex(VA::AbstractVectorOfArray, i::Union{Int,A
   function AbstractVectorOfArray_getindex_adjoint(Δ)
     Δ′ = zero(VA)
     Δ′[i,j...] = Δ
-    (Δ′, i,map(_ -> nothing, j)...)
+    @show Δ′
+    # (Δ′, i,map(_ -> nothing, j)...)  # surely that i is a bug?
+    (Δ′, nothing, map(_ -> nothing, j)...)
+    # (VectorOfArray(Δ′), nothing, map(_ -> nothing, j)...)
```
ChrisRackauckas, line 79: why would it not be this one?

src/zygote.jl
```diff
@@ -55,11 +73,13 @@ ZygoteRules.@adjoint function getindex(VA::AbstractVectorOfArray, i::Union{Int,A
   function AbstractVectorOfArray_getindex_adjoint(Δ)
     Δ′ = zero(VA)
     Δ′[i,j...] = Δ
-    (Δ′, i,map(_ -> nothing, j)...)
+    @show Δ′
```
ChrisRackauckas, line 76:
```suggestion
```

ChrisRackauckas:
Thanks! Running downstream tests to check it out.

src/zygote.jl
mcabbott: Sorry, very WIP!

src/zygote.jl
```diff
@@ -55,11 +73,13 @@ ZygoteRules.@adjoint function getindex(VA::AbstractVectorOfArray, i::Union{Int,A
   function AbstractVectorOfArray_getindex_adjoint(Δ)
     Δ′ = zero(VA)
     Δ′[i,j...] = Δ
-    (Δ′, i,map(_ -> nothing, j)...)
+    @show Δ′
+    # (Δ′, i,map(_ -> nothing, j)...)  # surely that i is a bug?
```
ChrisRackauckas, line 77: yeah that's a bug

src/zygote.jl
ChrisRackauckas: oh no worries, still worth the downstream tests

PR #169, opened 2021-09-29
Gradient definitions & supertypes for Zygote, continued

ChrisRackauckas:
Continues https://github.com/SciML/RecursiveArrayTools.jl/pull/168

@mcabbott

I didn't have the write permissions for some reason so continuing it here.


src/zygote.jl
```diff
@@ -10,7 +10,7 @@ function ChainRulesCore.rrule(::typeof(getindex),VA::AbstractVectorOfArray, indi
   function AbstractVectorOfArray_getindex_adjoint(Δ)
     Δ′ = zero(VA)
     Δ′[indices...] = Δ
-    (NoTangent(), Δ′, indices[1],map(_ -> NoTangent(), indices[2:end])...)
+    (NoTangent(), VectorOfArray(Δ′), indices[1],map(_ -> NoTangent(), indices[2:end])...)
```
mcabbott, line 13: I think this one wants nothing instead of `indices[1]` too:
```suggestion
    (NoTangent(), VectorOfArray(Δ′), map(_ -> NoTangent(), indices)...)
```

src/zygote.jl
```diff
@@ -43,10 +43,21 @@ function ChainRulesCore.rrule(::typeof(getproperty),A::ArrayPartition, s::Symbol
     A.x,literal_ArrayPartition_x_adjoint
 end
 
+# Define a new species of projection operator for this type:
+ChainRulesCore.ProjectTo(x::VectorOfArray) = ChainRulesCore.ProjectTo{VectorOfArray}()
+
+# Gradient from iteration will be e.g. Vector{Vector}, this makes it another AbstractMatrix
+(::ChainRulesCore.ProjectTo{VectorOfArray})(dx::AbstractVector{<:AbstractArray}) = VectorOfArray(dx)
+# Gradient from broadcasting will be another AbstractArray
+(::ChainRulesCore.ProjectTo{VectorOfArray})(dx::AbstractArray) = dx
```
mcabbott, line 52: I believe this may not be necessary at all. One thing I thought to test was whether iteration like this worked without it, but it does, it hits the `@adjoint getindex` rule:
```
julia> function iter(vofa)
       s = 0
       for a in vofa
         s += prod(a)
       end
       s
       end;

julia> gradient(iter, va)[1]
VectorOfArray{Float64,3}:
2-element Vector{Matrix{Float64}}:
 [0.007377548303139424 0.0014293014720444096 0.0004998127128840348; 0.0005414269337139141 0.0007721441834498009 0.0006559506948612249; 0.0042378737935180105 0.0006765914005991947 0.00045986425172967415]
 [0.002774992305796606 0.002978041675310144 0.004412709924140469; 0.0056425205408066285 0.005228088118453952 0.0036646150274027; 0.0036825199036535465 0.004902176341789764 0.045170987413739046]
```

src/zygote.jl
```diff
@@ -43,10 +43,21 @@ function ChainRulesCore.rrule(::typeof(getproperty),A::ArrayPartition, s::Symbol
...
+(::ChainRulesCore.ProjectTo{VectorOfArray})(dx::AbstractVector{<:AbstractArray}) = VectorOfArray(dx)
+# Gradient from broadcasting will be another AbstractArray
+(::ChainRulesCore.ProjectTo{VectorOfArray})(dx::AbstractArray) = dx
+
+# These rules duplicate the `rrule` methods above, because Zygote looks for an `@adjoint`
+# definition first, and finds its own before finding those.
+
 ZygoteRules.@adjoint function getindex(VA::AbstractVectorOfArray, i::Union{Int,AbstractArray{Int},CartesianIndex,Colon,BitArray,AbstractArray{Bool}})
   function AbstractVectorOfArray_getindex_adjoint(Δ)
     Δ′ = [ (i == j ? Δ : zero(x)) for (x,j) in zip(VA.u, 1:length(VA))]
```
mcabbott, line 59: Unrelated to the PR, but this seems like it allocates quite a bit, when iterating a VectorOfArray. I guess that using `Fill(0.0, size(Δ))` would often make `Δ′` have an abstract type?

ChrisRackauckas: Yes, it would make it an abstract type and sometimes hurt inference. Then we'd have to rely on union optimizations and pray.

ChrisRackauckas: Relying on union optimizations might be the right idea here though, I'll have to check.

PR #173, opened 2021-11-25
Provide base case for recursive eltype functions

baggepinnen:
Addresses https://github.com/SciML/ModelingToolkit.jl/issues/1353
The problem in the linked issue is not solved completely since the eltype infers to `Any`, but this PR prevents the stackoverflow.


ChrisRackauckas:
This won't actually solve the problem because then the solver will just fail somewhere else not being able to build `zero(Any)`. But it is probably good to stop the overflow anyways.

Issue #178, opened 2021-12-31
Indexing VectorOfArrays on GPU throws "scalar indexing" warning


ma-sadeghi:
```julia
x = CUDA.zeros(5)
y = VectorOfArray([x,x,x])
y[:,:]
```
The last line produces the following warning:
```julia
┌ Warning: Performing scalar indexing on task Task (runnable) @0x00007efcf315c010. Invocation of getindex resulted in scalar indexing of a GPU array. This is typically caused by calling an iterating
│ implementation of a method. Such implementations do not execute on the GPU, but very slowly on the CPU, and therefore are only permitted from the REPL for prototyping purposes. If you did intend to
│ index this array, annotate the caller with @allowscalar.
└ @ GPUArrays /home/anaconda/.julia/packages/GPUArrays/gkF6S/src/host/indexing.jl:56
5×3 Matrix{Float32}:
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
 0.0  0.0  0.0
```

ChrisRackauckas:
@ma-sadeghi try out https://github.com/SciML/RecursiveArrayTools.jl/pull/179 ?

ma-sadeghi:
@ChrisRackauckas Thanks for the quick PR! I just tired it out and sure enough, I don't see the warning anymore!

PR #179, opened 2022-01-02
Fix all colon dispatch for GPUs

ChrisRackauckas:
Fixes https://github.com/SciML/RecursiveArrayTools.jl/issues/178 and sets up GPU testing


ma-sadeghi (commented):
@ChrisRackauckas I think in the unit tests that you've added, if you add `CUDA.allowscalar(false)` after you import `CUDA`, then the warning turns into an actual error; might make more sense for testing purposes?

PR #181, opened 2022-01-04
Specialize getindex for indexing VectorOfArray with mask and colon

### LINKED ISSUE #180: Indexing VectorOfArrays using boolean masks throws warning
```julia
x = CUDA.rand(5, 3)
m = CUDA.rand(5) .> 0.5
x[m, :]  # this works just fine

u = VectorOfArray([x, x, x])
u[m, :]  # this throws warning/error
```

ma-sadeghi:
Fixes #180

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -78,7 +78,7 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractVectorOfArray{T, N},
   @assert length(J) == ndims(A.u[1])+1-ndims(I)
   @assert size(I) == size(A)[1:ndims(A)-length(J)]
   vecs = vec.(A.u)
-  return Base.getindex(Adapt.adapt(__parameterless_type(T),reshape(reduce(hcat,vecs),size(A.u[1])...,length(A.u))), I, J...)
+  return Adapt.adapt(__parameterless_type(T),reshape(reduce(hcat,vecs[I]),size(A.u[1])...,length(A.u)))
 end
 
 Base.@propagate_inbounds Base.getindex(A::AbstractDiffEqArray{T, N}, i::Int,::Colon) where {T, N} = [A.u[j][i] for j in 1:length(A)]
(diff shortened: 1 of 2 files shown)


ma-sadeghi:
Indexing with a mask for some dimensions (and not others) is still not supported, something like: `A[:, m, :, :]` for a 4D `VectorOfArray`. Didn't know how to fix that (I know how to deal with `A[:, m, :, :]` in particular, but don't know how to generalize it if we have multiple masks in between colons, not sure how practical this kind of indexing is, so maybe not even worth it!).

ChrisRackauckas (approved):

src/vector_of_array.jl
```diff
@@ -73,6 +73,14 @@ Base.@propagate_inbounds function Base.getindex(A::AbstractVectorOfArray{T, N},
   return Adapt.adapt(__parameterless_type(T),reshape(reduce(hcat,vecs),size(A.u[1])...,length(A.u)))
 end
 
+Base.@propagate_inbounds function Base.getindex(A::AbstractVectorOfArray{T, N},
+  I::AbstractArray{Bool},J::Colon...) where {T, N}
+  @assert length(J) == ndims(A.u[1])+1-ndims(I)
+  @assert size(I) == size(A)[1:ndims(A)-length(J)]
+  vecs = vec.(A.u)
+  return Base.getindex(Adapt.adapt(__parameterless_type(T),reshape(reduce(hcat,vecs),size(A.u[1])...,length(A.u))), I, J...)
```
ChrisRackauckas, line 81:
```suggestion
  return Adapt.adapt(__parameterless_type(T),reshape(reduce(hcat,vecs[I]),size(A.u[1])...,length(A.u)))
```

Seems like it would do it?

ma-sadeghi:
@ChrisRackauckas Thanks for your review. I just tried it out and I still get the scalar indexing error, which seems to be coming from this part: `reduce(hcat,vecs[I])`, and more specifically the `vecs[I]` part.

ChrisRackauckas:
Yeah it looks like the tests went brain-dead and didn't run that one, so I merged accidentally just seeing that all tests passed 😰 .

ChrisRackauckas:
https://github.com/SciML/RecursiveArrayTools.jl/pull/182

PR #182, opened 2022-01-04
Try another mask for GPU overload

ChrisRackauckas:


ma-sadeghi:
@ChrisRackauckas It seems that GPUArrays still complains about `vec.(A.u[I])`. The only workaround that I can think of is:
```julia
vecs = vec.(A.u)
return Base.getindex(Adapt.adapt(__parameterless_type(T),reduce(hcat,vecs),size(A.u[1])...,length(A.u))),
                     I, J...)
```
which is not elegant...

ma-sadeghi:
@ChrisRackauckas Yeah, we were missing the `reshape` bit. But, actually if you replace the `return` line with:
```julia
return A[repeat([:], length(size(A)))...][I,J...]
```
It's shorter and easier to read. I tested it locally and it works. Wanted to commit, but didn't have permission.

ChrisRackauckas:
Open a new PR.

PR #188, opened 2022-01-31
run tests on personal machines

vpuri3:
no tests were running on my personal machine earlier.
```julia
     Testing Running tests...
  0.000008 seconds
     Testing RecursiveArrayTools tests passed 

(RecursiveArrayTools) pkg> 
```

with this change, all the "light" tests will run on personal machines

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -19,13 +19,14 @@ end
 
 @time begin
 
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
This is incorrect, just all the group all is a better solution.

PR #189, opened 2022-01-31
remove functions that are predefined with `abstractarray` interface

vpuri3:
unless there's performance diff

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.jl.mem
 Manifest.toml
 .vscode
+*.swp # vim swap files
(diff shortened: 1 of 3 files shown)


vpuri3:
All `Core` tests passed on my machine

ChrisRackauckas:
What's the performance difference?

vpuri3:
there isn't any. i meant to say "remove those functions unless there is a performance difference for for a particular use case". I don't see any.

ChrisRackauckas:
Could you post some benchmarks?

mateuszbaran:
Here are some benchmark examples I care about in case someone wants to check performance impact: https://github.com/SciML/RecursiveArrayTools.jl/issues/160

vpuri3:
thanks will get to it

PR #205, opened 2022-05-24
Require ArrayInterfaceStaticArrays to fix downstream

ChrisRackauckas:


ChrisRackauckas:
@oshulz this is a bit of a tragedy. But there's a test that shows it's necessary, because of how this is implemented:

```julia
@generated function ArrayInterfaceCore.ismutable(::Type{<:ArrayPartition{T,S}}) where {T,S}
    res = all(ArrayInterfaceCore.ismutable, S.parameters)
    return :( $res )
end
```

which needs the dispatch for static arrays already defined for this to be resolved correctly, since it doesn't check the back edges.

oschulz:
I guess that brings us back to the question "what would we need in a StaticArraysCore"?

ChrisRackauckas:
Once again, this would only need `SArray` in there to know to mark it as `ismutable` false. So, the same little subset, nothing changed.

ChrisRackauckas:
IMO, first push to Base should probably be to get that concept in Base. I don't know how people could write code without, it's insane that it still doesn't exist.

PR #236, opened 2022-12-07
Add new interface for symbolic indexing

AayushSabharwal:
- DiffEqArray now stores the new `SymbolCache` struct, which defines and implements its interface to query symbols
- DiffEqArray supports symbolically indexing parameters, if provided

See https://github.com/SciML/ModelingToolkit.jl/pull/1988

All tests pass as-is, so this shouldn't be breaking. The constructors are modified to have the same interface but create the new structures

### DIFF
--- a/src/tabletraits.jl
+++ b/src/tabletraits.jl
@@ -7,12 +7,12 @@ function Tables.rows(A::AbstractDiffEqArray)
         N = length(A.u[1])
         names = [
             :timestamp,
-            (A.syms !== nothing ? (A.syms[i] for i in 1:N) :
+            (A.sc !== nothing && A.sc.syms !== nothing ? (A.sc.syms[i] for i in 1:N) :
              (Symbol("value", i) for i in 1:N))...,
         ]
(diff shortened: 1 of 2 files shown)


AayushSabharwal:
This gets a little annoying due to the compatibility with the old route:
```
DiffEqArray(u, t, sys, observed, p)
```
ends up constructing a `SymbolCache(sys, observed, nothing)` and puts `p` where `observed` should be.

test/symbolic_indexing_interface_test.jl
```diff
@@ -0,0 +1,15 @@
+using RecursiveArrayTools, Test
```
ChrisRackauckas, line 1: needs to be added to runtests.jl

AayushSabharwal: Oops 😅

ChrisRackauckas:
That's a fine wart for now. Open an issue to deprecate that in the next breaking update.

PR #244, opened 2023-01-25
Ambiguity errors and format

chriselrod:
@AayushSabharwal I'm trying to see if adding `::NTuple{N, Int}` is a valid signature on `DiffEqArray`.
Having `::NTuple{N}` in the third position is compatible with `NTuple{N,Symbol}`, resulting in ambiguities with the following two constructors.

This is an attempt to avoid that ambiguity, while being totally ignorant of what that third argument in the first constructor is actually supposed to be.


AayushSabharwal:
I'm not sure what it's for either, and don't remember seeing it used anywhere. A quick browse through the parts of SciMLBase I know also doesn't use it. Hopefully we should be good?

PR #247, opened 2023-02-06
Add ReverseDiff weak dep

ChrisRackauckas:
with backwards compatibility. Comes from SciMLSensitivity.jl back compat split.


src/RecursiveArrayTools.jl
```diff
@@ -42,6 +42,13 @@ function ChainRulesCore.rrule(T::Type{<:GPUArraysCore.AbstractGPUArray},
     T(xs), ȳ -> (NoTangent(), ȳ)
 end
 
+import Requires
+function __init__()
+    @static if !isdefined(Base, :get_extension)
```
chriselrod, line 47: Could move this check outside the function to avoid loading it.

oscardssmith (approved):

ext/RecursiveArrayToolsReverseDiffExt.jl
```diff
@@ -0,0 +1,15 @@
+module RecursiveArrayToolsReverseDiffExt
+
+function RecursiveArrayTools.recursivecopy!(b::AbstractArray{T, N},
+                                            a::AbstractArray{T2, N}) where {
+                                                                            T <:
+                                                                            Tracker.TrackedArray,
+                                                                            T2 <:
```
chriselrod, line 7: `Tracker.TrackedArray` probably isn't in scope?

ChrisRackauckas: lol...

PR #265, opened 2023-05-14
`issymbollike()` has been extracted into `SymbolicsBase` package (Fixes #264)

### LINKED ISSUE #264: Could `issymbollike` be extracted into a new package?
I'd like to see `Symbolics.jl` supported in [`Measurements.jl`](https://github.com/JuliaPhysics/Measurements.jl),
but for reasons, the `Num` type would need to be explicitly allowed
on the `Measurements.jl` side, which will require depending on
whatever package providing it. 

I've since found the `SciMLBase.issymbollike()` function,
which should answer the question i'm looking for,
but unfortunately, if i add `SciMLBase` as a dependency of `Measurements`,
to my surprise, we get a cyclic dependency, because `SciMLBase` exports `issymbollike`
from `RecursiveArrayTools`, which itself depends on `Measurements`...

So, can `issymbollike` be extracted into a new package?

Refs. https://github.com/JuliaPhysics/Measurements.jl/issues/142#issuecomment-1537170926
Refs. https://github.com/JuliaSymbolics/Symbolics.jl/issues/898

Thanks!

LebedevRI:
I'd like to see `Symbolics.jl` supported in [`Measurements.jl`](https://github.com/JuliaPhysics/Measurements.jl), but for reasons, the `Num` type would need to be explicitly allowed on the `Measurements.jl` side, which will require depending on whatever package providing it.

I've since found the `SciMLBase.issymbollike()` function, which should answer the question i'm looking for,
but unfortunately, if i add `SciMLBase` as a dependency of `Measurements`, to my surprise, we get a cyclic dependency,
because `SciMLBase` exports `issymbollike`
from `RecursiveArrayTools`, which itself depends on `Measurements`...

By extracting `issymbollike()` into a new package, on which `Measurements.jl` could depend,
we avoid dependency cycle.

This, obviously, requires said `SymbolicsBase.jl` to be made avaliable.
I've put a repo at https://github.com/LebedevRI/SymbolicsBase.jl,
but it needs to be in this org if this is to proceed.

Refs. https://github.com/JuliaPhysics/Measurements.jl/issues/142#issuecomment-1537170926
Refs. https://github.com/JuliaSymbolics/Symbolics.jl/issues/898

Thanks!


ChrisRackauckas:
Why would this be needed when there's https://github.com/SciML/SymbolicIndexingInterface.jl ? And BasicSymbolic as Unityper?

LebedevRI:
Ok, this is indeed not strictly needed, same effect can be mimicked via a `@require`.
Thank you!

PR #267, opened 2023-07-12
Make `lastindex(::AbstractVectorOfArray, i)` error when `i!=1`

oscardssmith:
This makes `A[:, end]` error for ragged `AbstractVectorOfArray` when the lengths are not equal. Before it could either index out of bounds or get non-last indices.


src/vector_of_array.jl
```diff
@@ -544,3 +544,10 @@ end
 end
 unpack_args_voa(i, args::Tuple{Any}) = (unpack_voa(args[1], i),)
 unpack_args_voa(::Any, args::Tuple{}) = ()
+
+function Base.lastindex(A::AbstractVectorOfArray, i=1)
+    i == 1 && return length(A)
+    len1 = lastindex(A[1], i-1)
+    all(x->isequal(lastindex(x, i-1), len1), A) && return len1
+    throw(ArgumentError("Can not get lastindex of an AbstractVectorOfArray for dimensions other than 1"))
```
ChrisRackauckas, line 552:
```suggestion
    throw(ArgumentError("`end` is not defined for AbstractVectorOfArray types when the arrays have different sizes. Either enforce that the arrays are all similarly sized or directly define the index value. 
    
Note that a common reason for this error in the SciMLSolution context comes from adaptively sized models. Using arguments like `saveat` can enforce that an ensemble of solutions are all saved at the same time points and thus all have the same size. Similarly, if an ODE is solved with adaptivity in the size of the system (such as for adaptive grids) then this error can refer to the changing size of the ODE through the time series."))
```

PR #276, opened 2023-09-04
Remove type constraints for copy of ArrayPartition

### LINKED ISSUE #275: Loosen type constraints for ArrayPartition copy?
I tried to make use of an `ArrayPartition` to prevent [Arrow.jl](https://github.com/apache/arrow-julia) arrays from being loaded into ram when concatenating tables.

The problem is that as soon as you want to actually materialise an array to do some operation the type constraints of [`copy`](https://github.com/SciML/RecursiveArrayTools.jl/blob/6e8600fe96a10112a59e146051d29399370a09b7/src/array_partition.jl#L84) are violated since a copied arrow array becomes a normal array (which I guess is due to the original array being a special type which is mmaped to disk).

I understand this is not the intended use case of this package, but I wonder if something like
```julia
function Base.copy(A::ArrayPartition{T}) where T
  newx = copy.(A.x)
  ArrayPartition{T, typeof(newx)}(newx)
end
```
would work instead of current implementation? This is still type stable (as long as `copy.(A.x)` is type stable) so it should not have any performance disadvantages.

DrChainsaw:
Fix #275 

Seems like it was ok to just remove all type constraints. I did a quick check with a dummy array type and the performance was identical to current main.

~~I tried to do the same thing for `zero`, but then it would no longer infer (even with the same construction as in the example in #275). Since I don't seem to need it for my use case I just left it as is.~~

Also did the same for `zero` since it had the same type of issue.

The tests might seem a bit silly. I put it in a separate commit in case you don't want it.

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -81,7 +81,7 @@ function Base.similar(A::ArrayPartition, ::Type{T}, ::Type{S}, R::DataType...) w
     ArrayPartition(f, N)
 end
 
-Base.copy(A::ArrayPartition{T, S}) where {T, S} = ArrayPartition{T, S}(copy.(A.x))
+Base.copy(A::ArrayPartition) = ArrayPartition(copy.(A.x))
 
 ## zeros
 Base.zero(A::ArrayPartition{T, S}) where {T, S} = ArrayPartition{T, S}(zero.(A.x))
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -81,10 +81,10 @@ function Base.similar(A::ArrayPartition, ::Type{T}, ::Type{S}, R::DataType...) w
     ArrayPartition(f, N)
 end
 
-Base.copy(A::ArrayPartition) = ArrayPartition(copy.(A.x))
+Base.copy(A::ArrayPartition) = ArrayPartition(map(copy, A.x))
 
 ## zeros
-Base.zero(A::ArrayPartition{T, S}) where {T, S} = ArrayPartition{T, S}(zero.(A.x))
+Base.zero(A::ArrayPartition) = ArrayPartition(map(zero, A.x))
 # ignore dims since array partitions are vectors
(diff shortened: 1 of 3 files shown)


DrChainsaw:
Just realized current version does not infer correctly for nested ArrayPartitions. Changing the broadcast to map seems to fix this. Running tests locally atm to verify that nothing else breaks from this.

ChrisRackauckas (approved):

DrChainsaw:
I tried running the formatter like this:
```julia
julia> using JuliaFormatter

julia> format_file("src/array_partition.jl"; style=SciMLStyle())
false
```

But it did alot of changes to the file and did not touch the lines changed by this PR so I didn't commit it. I also tried with the default style with the same result. 

Is there some other way to run it or is the file unformatted on current main?

ChrisRackauckas:
The format changed. I wouldn't worry about that. I'm going to fix formatter things in the near future.

https://github.com/SciML/RecursiveArrayTools.jl/actions/runs/6076983375/job/16487010044?pr=276#step:6:414

It seems you branched off of an earlier master that fixed the Project.toml form though. Can you try rebasing?

DrChainsaw:
> It seems you branched off of an earlier master that fixed the Project.toml form though. Can you try rebasing?

Hmmm, I can't seem to find any missing commit on the branch. Below is the last one which is not mine.

```
commit 6e8600fe96a10112a59e146051d29399370a09b7 (origin/master, origin/HEAD, master)   
Merge: d06ecb8 727a4b6
Author: Christopher Rackauckas <accounts@chrisrackauckas.com>
Date:   Mon Aug 21 19:04:16 2023 +0200

    Merge pull request #272 from visr/visr-patch-1
    
    fix UndefVarError: `_throw_dmrs` not defined
```

ChrisRackauckas:
Seems like it was just an aqua issue on v1.6 so I disabled it for earlier julia

PR #279, opened 2023-09-12
Support nesting Unitful and MonteCarloMeasurements

ChrisRackauckas:
Fixes https://discourse.julialang.org/t/error-when-using-unitful-and-montecarlomeasurements-with-differentialequations/103464


baggepinnen:
Did the tests run? I think there's a typo in the import statement, importing Measurements instead of MonteCarloMeasurements.jl

ChrisRackauckas:
yeah oops forgot to add the test file https://github.com/SciML/RecursiveArrayTools.jl/pull/280

PR #287, opened 2023-10-03
simplify issymbollike

oscardssmith:
I'm not 100% sure this is correct, but it is a lot cleaner. It's also somewhat unclear to me why the Symbolics.jl types are getting hacked around here rather than having Symbolics overload `issymbollike` (possibly with a package extension).

### DIFF
--- a/ext/RecursiveArrayToolsSymbolicsExt.jl
+++ b/ext/RecursiveArrayToolsSymbolicsExt.jl
@@ -3,6 +3,6 @@ module RecursiveArrayToolsSymbolicsExt
 import RecursiveArrayTools
 isdefined(Base, :get_extension) ? (import Symbolics) : (import ..Symbolics)
 
-RecursiveArrayTools.issymbollike(::Symbolics.Num) = true
+RecursiveArrayTools.issymbollike(::Union{Symbolics.BasicSymbolic,Symbolics.Num}) = true
 
 end
--- a/ext/RecursiveArrayToolsSymbolicsExt.jl
+++ b/ext/RecursiveArrayToolsSymbolicsExt.jl
@@ -1,0 +1,8 @@
+module RecursiveArrayToolsSymbolicsExt
+
+import RecursiveArrayTools
+isdefined(Base, :get_extension) ? (import Symbolics) : (import ..Symbolics)
+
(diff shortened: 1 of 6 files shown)


src/vector_of_array.jl
```diff
@@ -64,16 +64,10 @@ end
 struct AllObserved
 end
 
-Base.@pure __parameterless_type(T) = Base.typename(T).wrapper
```
staticfloat, line 67: This method is used on line 197, caught by the GPU tests.

src/vector_of_array.jl
```diff
@@ -64,16 +64,10 @@ end
...
-@generated function issymbollike(x)
-    x <: Union{Symbol, AllObserved} && return quote true end
-    ss = ["Operation", "Variable", "Sym", "Num", "Term"]
-    s = string(Symbol(__parameterless_type(x)))
-    bool = any(x -> occursin(x, s), ss) 
-    quote 
-        $bool
-    end
+function issymbollike(x)
+    x isa Union{Symbol, AllObserved} && return true
```
ChrisRackauckas, line 68: Just dispatch in Symbolics in an extension

oscardssmith: what should the behavior be in pre 1.9?

ext/RecursiveArrayToolsSymbolicsExt.jl
```diff
@@ -0,0 +1,8 @@
+module RecursiveArrayToolsSymbolicsExt
+
+import RecursiveArrayTools
+isdefined(Base, :get_extension) ? (import Symbolics) : (import ..Symbolics)
+
+RecursiveArrayTools.issymbollike(::Symbolics.Num) = true
```
ChrisRackauckas, line 6:
```suggestion
RecursiveArrayTools.issymbollike(::Union{Symbolics.BasicSymbolic,Symbolics.Num}) = true
```

ChrisRackauckas (approved):

oscardssmith:
oh this is somewhat annoying this is actually a recursive dependency. We need the package extension to live in Symbolics, but that will cause problems since the versions then have to be coordinated.

PR #289, opened 2023-10-18
Replace version 2.39.0 with version 3.0

oscardssmith:
https://github.com/SciML/RecursiveArrayTools.jl/pull/287 was actually a breaking change


ChrisRackauckas:
There's 100 dependents. Can we also do the complete cleanup of the symbolic indexing interface at the same time? @AayushSabharwal

ChrisRackauckas:
At least let's bound for safety 👮

PR #290, opened 2023-10-23
feat!: rework AbstractVectorOfArray, use new SymbolicIndexingInterface

### LINKED ISSUE #237: [Breaking] Update DiffEqArray constructors
To avoid breaking changes in #236, the constructors for DiffEqArray had been modified to work the same way they did before. However, this makes it so that trying to construct a DiffEqArray with a system requires specifying type parameters:

```julia
DiffEqArray(u, t, syms, indepsym, observed, p) # works as it did before
DiffEqArray(u, t, sys, observed, p) # interprets sys as syms, observed as indepsym and p and observed
DiffEqArray{..., typeof(sys), typeof(observed), typeof(p)}(u, t, sys, observed, p) # works
```

Also, to specify paramsyms (without a system)
```julia
sc = SymbolCache(syms, indepsym, paramsyms)
DiffEqArray{..., typeof(sc), typeof(observed), typeof(p)}(u, t, sc, observed, p)
```

In the next breaking change, constructors would need to be updated to make the API smoother

AayushSabharwal:
- Implemented new SII
- `AbstractVectorOfArray` doesn't subtype `AbstractArray`
- Removed `issymbollike`

Closes #237


AayushSabharwal:
Removing the `AbstractArray` subtype has broken broadcasting, which I'm trying to figure out

AayushSabharwal:
I've tried to make `AbstractVectorOfArray` work the same as it did before without subtyping `AbstractArray`. I'm not particularly happy with all of the changes made, so I'm open to input

YingboMa (approved):

src/array_partition.jl
```diff
@@ -256,26 +248,18 @@ end
...
-    setindex!(A::ArrayPartition, v, i::Int, j...)
-
-Set the entry at index `j...` of the `i`th partition of `A` to `v`.
-"""
-Base.@propagate_inbounds function Base.setindex!(A::ArrayPartition, v, i::Int, j...)
-    @boundscheck 0 < i <= length(A.x) || throw(BoundsError(A.x, i))
-    @inbounds b = A.x[i]
-    @boundscheck checkbounds(b, j...)
-    @inbounds b[j...] = v
+function Base._maybe_reshape(::IndexCartesian,
```
ChrisRackauckas, line 259: why was this removed?

src/vector_of_array.jl
```diff
@@ -119,198 +138,255 @@ function VectorOfArray(vec::AbstractVector{VT}) where {T, N, VT <: AbstractArray
...
-end
-Base.@propagate_inbounds function Base.getindex(A::AbstractDiffEqArray{T, N},
-                                                I::Union{Int, AbstractArray{Int},
-                                                         CartesianIndex, Colon, BitArray,
-                                                         AbstractArray{Bool}}...) where {T,
-                                                                                         N}
-    RecursiveArrayTools.VectorOfArray(A.u)[I...]
+
+function Base.lastindex(VA::AbstractVectorOfArray)
+    Base.depwarn("Linear indexing of `AbstractVectorOfArray` is deprecated", :lastindex)
```
ChrisRackauckas, line 220:
```suggestion
    Base.depwarn("Linear indexing of `AbstractVectorOfArray` is deprecated. Change `A[i]` to `A.u[i]` ", :lastindex)
```

src/array_partition.jl
AayushSabharwal: In 1.10 the internal implementation of `\` changed, and it indexes `ArrayPartition` with two indexes (the last of which is always `1`. That ends up calling this dispatch, and segfaults since the bounds checks are ignore (`@inbounds`). Adding methods for `Base.:\` and `ldiv!` ends up cascading into more methods for `mul!` and so on, which all need to only be defined for 1.10

PR #298, opened 2023-12-15
[do not merge] Rerun CI

AayushSabharwal:
## Checklist


src/vector_of_array.jl
```diff
@@ -510,7 +510,7 @@ function Base.CartesianIndices(VA::AbstractVectorOfArray)
 end
 
 # Tools for creating similar objects
-Base.eltype(::VectorOfArray{T}) where {T} = T
+Base.eltype(::Type{<:AbstractVectorOfArray{T}}) where {T} = T
```
ChrisRackauckas, line 513: don't we want both?

AayushSabharwal: `eltype` docs say to define the latter. There already is a dispatch in Base for `eltype(x) = eltype(typeof(x))`

AayushSabharwal:
Tests are fixed by SciMLBase and DiffEqBase PRs

AayushSabharwal:
In hindsight, I guess CI targeting 1.9 specifically should be removed until 1.10 is out. Seems pointless using CI hours on identical tests.

src/vector_of_array.jl
ChrisRackauckas: 👍

src/array_partition.jl
```diff
@@ -1,3 +1,5 @@
+using SparseArrays
```
ChrisRackauckas, line 1: wrong location for this.

AayushSabharwal: I'll fix this and drop 1.9-specific CI once tests pass. It's absurd that they're taking over half an hour

AayushSabharwal:
There is something wrong with CI

src/array_partition.jl
AayushSabharwal: Fixed

AayushSabharwal:
I'm working on fixing compat

PR #325, opened 2024-01-04
fix: mapreduce type stability, VoA broadcast adjoints

### LINKED ISSUE #323: Type instability when iterating nested recursive arrays
I noticed that the following is type unstable (on v3.3.3 and julia 1.10.0)

```julia
using RecursiveArrayTools

sum(ArrayPartition(ArrayPartition(zeros(4,4)))) # ::Any

sum(VectorOfArray([zeros(4), VectorOfArray([zeros(4,4)])])) # ::Any
```

Curiously, all of the following are stable.
```julia
sum(ArrayPartition(ArrayPartition(zeros(4)))) # ::Float64
sum(ArrayPartition(zeros(4,4))) # ::Float64

sum(VectorOfArray([VectorOfArray([zeros(4,4)])])) # ::Float64
```

The other recursive array types seem to be affected as well. Not sure if getting these stable is in-scope, but it would be super helpful.

AayushSabharwal:
Close #323 

## Checklist

### DIFF
--- a/ext/RecursiveArrayToolsZygoteExt.jl
+++ b/ext/RecursiveArrayToolsZygoteExt.jl
@@ -100,6 +100,11 @@ end
     end
 end
 
+@adjoint function Base.copy(u::VectorOfArray)
+    copy(u),
+    y ->  (copy(y),)
+end
+
 @adjoint function DiffEqArray(u, t)
     DiffEqArray(u, t),
     y -> begin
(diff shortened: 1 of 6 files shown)


src/vector_of_array.jl
```diff
@@ -595,8 +604,14 @@ function Base.convert(::Type{Array}, VA::AbstractVectorOfArray)
 end
 
 # statistics
-@inline Base.sum(f, VA::AbstractVectorOfArray) = sum(f, Array(VA))
-@inline Base.sum(VA::AbstractVectorOfArray; kwargs...) = sum(Array(VA); kwargs...)
+@inline Base.sum(VA::AbstractVectorOfArray; kwargs...) = sum(identity, VA; kwargs...)
+@inline function Base.sum(f, VA::AbstractVectorOfArray; kwargs...)
+    if hasproperty(kwargs, :dims)
+        sum(Array(VA); kwargs...)
```
ChrisRackauckas, line 610: This can be a potentially big allocation

src/vector_of_array.jl
```diff
@@ -638,8 +653,9 @@ end
 end
 
 Base.map(f, A::RecursiveArrayTools.AbstractVectorOfArray) = map(f, A.u)
+
 function Base.mapreduce(f, op, A::AbstractVectorOfArray)
```
ChrisRackauckas, line 657: probably need to have an `init`

src/vector_of_array.jl
AayushSabharwal: That was the default implementation, but I just realized two things:
- that method is probably not necessary
- mapreduce is implemented wrong

AayushSabharwal: Fixed in https://github.com/SciML/RecursiveArrayTools.jl/pull/325/commits/bdb58305372db57d38877e48157859e3137f7ccc

src/vector_of_array.jl
AayushSabharwal: Fixed in https://github.com/SciML/RecursiveArrayTools.jl/pull/325/commits/bdb58305372db57d38877e48157859e3137f7ccc

ext/RecursiveArrayToolsZygoteExt.jl
```diff
@@ -117,19 +122,122 @@ end
...
+    broadcast(+, x, y), ȳ -> (nothing, map(x -> Zygote.unbroadcast(x, ȳ), (x, y))...)
+end
+@adjoint function Broadcast.broadcasted(::typeof(+), x::Zygote.Numeric, y::AbstractVectorOfArray)
+    broadcast(+, x, y), ȳ -> (nothing, map(x -> Zygote.unbroadcast(x, ȳ), (x, y))...)
+end
+
+_minus(Δ) = .-Δ
+_minus(::Nothing) = nothing
+
+@adjoint function Broadcast.broadcasted(::typeof(-), x::AbstractVectorOfArray, y::Union{AbstractVectorOfArray, Zygote.Numeric})
```
ChrisRackauckas, line 152: you're never supposed to define specific broadcasts like this, and I don't think digging into Zygote's broadcast system is the answer. It just relies on the Julia-level broadcast

ChrisRackauckas:
This will need a real fix, but let's at least get downstream tests passing.

PR #331, opened 2024-01-10
fix: remove linear indexing from getindex adjoints

AayushSabharwal:
## Checklist


ChrisRackauckas:
It doesn't have an underlying tuple, it's an array of arrays. But yes, that is probably not right, it probably should recurse running adapt.

ChrisRackauckas:
do you have a test case that is sensitive to this?

ChrisRackauckas:
I made https://github.com/SciML/RecursiveArrayTools.jl/pull/434, but the tricky thing is always that GPUArrays cannot have other GPUArrays, so the structure would need to be Array of GPUArrays and I think VectorOfArray works out in that case?

PR #355, opened 2024-02-19
Update vector_of_array.jl

### LINKED ISSUE #354: `getindex` with slice breaks on size 1 arrays
MWE:

```julia
julia> using SciMLBase

julia> EnsembleSolution(ones(2), 0.0, false, nothing)[:]
2-element Vector{Float64}:
 1.0
 1.0

julia> EnsembleSolution(ones(1), 0.0, false, nothing)[:]
ERROR: MethodError: no method matching reshape(::Float64, ::Int64)

Closest candidates are:
  reshape(::RecursiveArrayTools.AbstractVectorOfArray, ::Any...)
   @ RecursiveArrayTools ~/.julia/packages/RecursiveArrayTools/kqKNI/src/vector_of_array.jl:687
  reshape(::FillArrays.AbstractFill, ::Union{Colon, Integer}...)
   @ FillArrays ~/.julia/packages/FillArrays/oXkMk/src/FillArrays.jl:260
  reshape(::AbstractArray, ::Int64...)
   @ Base reshapedarray.jl:117
  ...

Stacktrace:
 [1] _getindex
   @ ~/.julia/packages/RecursiveArrayTools/kqKNI/src/vector_of_array.jl:278 [inlined]
 [2] getindex(::EnsembleSolution{Float64, 1, Vector{Float64}}, ::Function)
   @ RecursiveArrayTools ~/.julia/packages/RecursiveArrayTools/kqKNI/src/vector_of_array.jl:393
 [3] top-level scope
   @ REPL[13]:1
```

This is because in the following line `reduce(hcat, vecs)` is not type stable: https://github.com/SciML/RecursiveArrayTools.jl/blob/87337d8e13556f5c8877b5a49705c5cce5f08699/src/vector_of_array.jl#L279

However, `Array` seems to work fine:
```julia
julia> Array(EnsembleSolution(ones(1), 0.0, false, nothing))
1-element Vector{Float64}:
 1.0
```
so the `getindex(A, ::Colon...)` method should probably get replaced with `Array`

lxvm:
Fix #354

## Checklist

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -296,13 +296,7 @@ __parameterless_type(T) = Base.typename(T).wrapper
 Base.@propagate_inbounds function _getindex(A::AbstractVectorOfArray{T, N},
         ::NotSymbolic, I::Colon...) where {T, N}
     @assert length(I) == ndims(A.u[1]) + 1
-    vecs = if N == 1
-        A.u
-    else
-        vec.(A.u)
-    end
-    return Adapt.adapt(__parameterless_type(T),
-        reshape(reduce(hcat, vecs), size(A.u[1])..., length(A.u)))
+    return Array(A)
 end
 
 Base.@propagate_inbounds function _getindex(A::AbstractVectorOfArray{T, N},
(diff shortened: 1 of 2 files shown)


AayushSabharwal (changes requested):
Copying isn't really an issue here, since we can't avoid it. However, I'd recommend using the `Adapt.adapt(...)` wrapper instead of `Array`. That way the output type matches the type of the contained arrays.

AayushSabharwal:
I think it copies, and anyway `Array(::VectorOfArray)` will copy anyway?

PR #357, opened 2024-02-20
Add support for `VectorOfArray(AbstractArray{<:AbstractArray})`

jlchan:
Addresses https://github.com/SciML/OrdinaryDiffEq.jl/pull/2087#issuecomment-1934232736

PR #359, opened 2024-02-21
Specialize `Base.similar` for `VectorOfArray` with multidimensional parent

jlchan:
This PR specializes `Base.similar` for `VectorOfArray` with a homogeneous (e.g., all elements are the same type) multidimensional parent array (introduced in https://github.com/SciML/RecursiveArrayTools.jl/pull/357). This ensures that `similar(VectorOfArray(array)).u` is similar to `array.u` where `array` is a homogeneous multi-dimensional array.

This PR also adds `Base.parent` for `VectorOfArray` (`parent(vec)` just returns `vec.u`)


jlchan:
@ranocha I think with these changes, we can deal with multidimensional arrays of `SVector`

For example, the following works for me under this PR:
```julia
using StaticArrays
using RecursiveArrayTools
using OrdinaryDiffEq

function rhs!(duu::VectorOfArray, uu::VectorOfArray, p, t)
    du = parent(duu)
    u = parent(uu)
    du .= u
end

u = fill(SVector{2}(ones(2)), 2, 3)
ode = ODEProblem(rhs!, VectorOfArray(u), (0.0, 1.0))
sol = solve(ode, Tsit5())
```

src/vector_of_array.jl
```diff
@@ -721,6 +724,18 @@ end
     VectorOfArray([similar(VA[:, i], T) for i in eachindex(VA.u)])
 end
 
+# for VectorOfArray with multi-dimensional parent arrays of arrays where all elements are the same type
+@inline function Base.similar(vec::VectorOfArray{
```
ChrisRackauckas, line 728:
```suggestion
function Base.similar(vec::VectorOfArray{
```

Is that well-tested? What's the reason for overriding the inlining calculation?

ChrisRackauckas:
Add that as a regression test in the downstream portion (https://github.com/SciML/RecursiveArrayTools.jl/blob/master/test/upstream.jl which is for some reason named upstream we should fix that and move it to the other test group https://github.com/SciML/RecursiveArrayTools.jl/blob/master/test/runtests.jl#L63

src/vector_of_array.jl
jlchan: Yeah, it's tested. 

I just copied the `@inline` from before, I didn't think about it overriding the other inlining.

jlchan: Thanks!

ChrisRackauckas (approved):

test/downstream/odesolve.jl
```diff
@@ -49,3 +49,14 @@ end
...
         (0.0, 1.0)),
     Rodas5()).retcode == ReturnCode.Success
+
+function rhs!(duu::VectorOfArray, uu::VectorOfArray, p, t)
+    du = parent(duu)
+    u = parent(uu)
+    du .= u
+end
+
+u = fill(SVector{2}(ones(2)), 2, 3)
```
jlchan, line 59:
```suggestion
u = fill(ones(2), 2, 3)
```
Would need to add `StaticArrays` as dependency otherwise.

jlchan: Nvm, StaticArrays is already a dependency

ChrisRackauckas: ohh... I'll just revert the downstream dep

ChrisRackauckas: wait no, it's only a core dep, not a StaticArrays dep, so this is right.

Issue #373, opened 2024-05-04
Broadcast over `VectorOfArray` doesn't respect dimensions of non-vector parent arrays


jlchan:
### **Is your feature request related to a problem? Please describe.**

`VectorOfArray` was extended to multi-dimensional parent arrays in https://github.com/SciML/RecursiveArrayTools.jl/pull/357. However, these arrays don't work with all solvers in OrdinaryDiffEq.jl (for example, `SSPRK43`). 

The error is in `integrator = init(ode, SSPRK43())`, specifically this line: https://github.com/SciML/OrdinaryDiffEq.jl/blob/1f2e058c14d2a5c997bf7d5f055cebfb406cbc1b/src/initdt.jl#L51. An MWE is
```julia
using OrdinaryDiffEq, RecursiveArrayTools

function rhs!(duu::VectorOfArray, uu::VectorOfArray, p, t)
    du = parent(duu)
    u = parent(uu)
    du .= u
end

u = fill(SVector{2}(ones(2)), 2, 3)
ode = ODEProblem(rhs!, VectorOfArray(u), (0.0, 1.0))
integrator = init(ode, SSPRK43())
```
The manually truncated stacktrace is
```julia
ERROR: DimensionMismatch: array could not be broadcast to match destination
Stacktrace:
  [1] check_broadcast_shape
    @ ./broadcast.jl:579 [inlined]
  [2] check_broadcast_axes
    @ ./broadcast.jl:582 [inlined]
  [3] instantiate
    @ ./broadcast.jl:309 [inlined]
  [4] materialize!
    @ ./broadcast.jl:914 [inlined]
  [5] materialize!
    @ ./broadcast.jl:911 [inlined]
  [6] rhs!(duu::VectorOfArray{…}, uu::VectorOfArray{…}, p::SciMLBase.NullParameters, t::Float64)
    @ Main ./Untitled-1:7
  [7] (::ODEFunction{…})(::VectorOfArray{…}, ::Vararg{…})
    @ SciMLBase ~/.julia/packages/SciMLBase/hSv8d/src/scimlfunctions.jl:2296 [inlined]
[...]

PR #374, opened 2024-05-06
Fix multi-dimensional `VectorOfArray` broadcast

jlchan:
Addresses https://github.com/SciML/RecursiveArrayTools.jl/issues/373

The fix just involves modifying `Base.copy` for `VectorOfArray` with multi-dimensional parent arrays. This could replace the old `Base.copy` implementation, but I left the old implementation there just in case it changes behavior upstream.

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -29,6 +29,7 @@ returns a vector of the series for each component, that is, `A[i,:]` for each `i
 A plot recipe is provided, which plots the `A[i,:]` series.
 
 There is also support for `VectorOfArray` constructed from multi-dimensional arrays
+
 ```julia
 VectorOfArray(u::AbstractArray{AT}) where {T, N, AT <: AbstractArray{T, N}}
 ```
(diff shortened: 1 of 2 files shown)


src/vector_of_array.jl
```diff
@@ -840,12 +840,37 @@ end
...
+        # this is the default behavior in v3.15.0
+        N = narrays(bc)
+        return VectorOfArray(map(1:N) do i
+            copy(unpack_voa(bc, i))
+        end)
+    else # if parent isa AbstractArray            
+        return VectorOfArray(map(enumerate(Iterators.product(axes(parent)...))) do (i, _)
+            copy(unpack_voa(bc, i))
+        end)
+    end
```
jlchan, line 873: Side note, this entire block can be replaced by the more general second branch 
```julia
return VectorOfArray(map(enumerate(Iterators.product(axes(parent)...))) do (i, _)
    copy(unpack_voa(bc, i))
end)
```
since `enumerate(Iterators.product(axes(parent)...))) do (i, _)` basically recovers `map(1:N) do i` for an `AbstractVector`. 

I left the old behavior in since I wasn't sure if it would break something upstream or change performance.

src/vector_of_array.jl
```diff
@@ -719,7 +718,7 @@ end
 # for VectorOfArray with multi-dimensional parent arrays of arrays where all elements are the same type
 function Base.similar(vec::VectorOfArray{
         T, N, AT}) where {T, N, AT <: AbstractArray{<:AbstractArray{T}}}
-    return VectorOfArray(similar(Base.parent(vec)))
+    return VectorOfArray(similar.(Base.parent(vec)))
```
jlchan, line 721: Broadcasting `similar` here avoids some `#undef` issues. 

Side note: this implementation could replace these 3 functions https://github.com/SciML/RecursiveArrayTools.jl/blob/eacbe3fa275440ea863db85e63d627dd103205d6/src/vector_of_array.jl#L715-L729

However, doing so breaks these tests because `similar` can't be called on `<:Real` values: https://github.com/SciML/RecursiveArrayTools.jl/blob/eacbe3fa275440ea863db85e63d627dd103205d6/test/basic_indexing.jl#L221-L229
These tests don't seem correct to me; the parent array `x` isn't a recursive array (`eltype(x)` is not `<:AbstractArray`), so to me it doesn't make sense to test this broadcast edge case. Maybe I'm missing something though.

jlchan:
Applying formatting introduces a bunch of spurious file changes so I'm skipping it for now.

src/vector_of_array.jl
jlchan: Side note: this implementation could replace these 3 functions https://github.com/SciML/RecursiveArrayTools.jl/blob/eacbe3fa275440ea863db85e63d627dd103205d6/src/vector_of_array.jl#L715-L729

However, doing so breaks these tests because `similar` can't be called on `<:Real` values: https://github.com/SciML/RecursiveArrayTools.jl/blob/eacbe3fa275440ea863db85e63d627dd103205d6/test/basic_indexing.jl#L221-L229
These tests don't seem correct to me; the parent array `x` doesn't have underlying data structure `Vector{AbstractArray{T}}`, so it shouldn't be a valid edge case. Maybe I'm missing something?

ChrisRackauckas: Numbers are also allowed. That's used a lot downstream.

ChrisRackauckas: We can special-case that case where the elements are numbers, but we cannot throw it out.

ChrisRackauckas (approved):

jlchan:
Thanks for the quick review!

PR #376, opened 2024-05-14
Device agnostic convert

avik-pal:
Needs testing.

Basically we can now use `convert(AbstractArray, x)` to do a device agnostic convert.

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -135,6 +135,9 @@ function Base.Array{U}(VA::AbstractVectorOfArray) where {U}
     vecs = vec.(VA.u)
     Array(reshape(reduce(hcat, vecs), size(VA.u[1])..., length(VA.u)))
 end
+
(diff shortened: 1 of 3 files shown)


src/vector_of_array.jl
```diff
@@ -135,6 +135,12 @@ function Base.Array{U}(VA::AbstractVectorOfArray) where {U}
     vecs = vec.(VA.u)
     Array(reshape(reduce(hcat, vecs), size(VA.u[1])..., length(VA.u)))
 end
+
+function Base.convert(::Type{AbstractArray}, VA::AbstractVectorOfArray)
+    vecs = vec.(VA.u)
+    return reshape(stack(vecs), size(VA.u[1])..., length(VA.u))
```
ChrisRackauckas, line 141: why is the reshape necessary after a stack? I thought that function was supposed to take care of that part?

avik-pal: my bad. fixed

PR #377, opened 2024-05-16
Remove projection rule for `VectorOfArray`

DhairyaLGandhi:
## Checklist

## Additional context

The current rule does not account for the various Tangent types that can be seen when working with downstream packages (specifically note AD'ing through observable functions via SII). It is much better to remove the rule and let AD take care of picking out the right fields etc. If its preferred to retain the rule (to get the VectorOfArray type returned in the reverse pass), then we will need to handle that elsewhere. For now, let's check for breakages in CI.

### DIFF
--- a/ext/RecursiveArrayToolsZygoteExt.jl
+++ b/ext/RecursiveArrayToolsZygoteExt.jl
@@ -139,19 +139,19 @@ end
     view(A, I...), view_adjoint
 end
 
-function ChainRulesCore.ProjectTo(a::AbstractVectorOfArray)
-    ChainRulesCore.ProjectTo{VectorOfArray}((sz = size(a)))
-end
-
-function (p::ChainRulesCore.ProjectTo{VectorOfArray})(x::Union{
-        AbstractArray, AbstractVectorOfArray})
-    if eltype(x) <: Number
-        arr = reshape(x, p.sz)
-        return VectorOfArray([arr[:, i] for i in 1:p.sz[end]])
-    elseif eltype(x) <: AbstractArray
-        return VectorOfArray(x)
-    end
-end
+# function ChainRulesCore.ProjectTo(a::AbstractVectorOfArray)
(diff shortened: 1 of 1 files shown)


ext/RecursiveArrayToolsZygoteExt.jl
```diff
@@ -139,19 +139,19 @@ end
...
-function (p::ChainRulesCore.ProjectTo{VectorOfArray})(x::Union{
-        AbstractArray, AbstractVectorOfArray})
-    if eltype(x) <: Number
-        arr = reshape(x, p.sz)
-        return VectorOfArray([arr[:, i] for i in 1:p.sz[end]])
-    elseif eltype(x) <: AbstractArray
-        return VectorOfArray(x)
-    end
-end
+# function ChainRulesCore.ProjectTo(a::AbstractVectorOfArray)
```
ChrisRackauckas, line 142: Please don't comment out code like this. You've done it in a few PRs. Just remove the code if it's to be removed.

ext/RecursiveArrayToolsZygoteExt.jl
```diff
@@ -139,19 +139,19 @@ end
...
+# 
+# function (p::ChainRulesCore.ProjectTo{VectorOfArray})(x::Union{
+#         AbstractArray, AbstractVectorOfArray})
+#     if eltype(x) <: Number
+#         arr = reshape(x, p.sz)
+#         return VectorOfArray([arr[:, i] for i in 1:p.sz[end]])
+#     elseif eltype(x) <: AbstractArray
+#         return VectorOfArray(x)
+#     end
+# end
```
ChrisRackauckas, line 154:
```suggestion
```

ext/RecursiveArrayToolsZygoteExt.jl
DhairyaLGandhi: Sure, will do.

PR #385, opened 2024-06-05
chore: set indexing style

DhairyaLGandhi:
## Checklist

## Additional context

ReverseDiff restricts array types to `IndexLinear()` indexing style as written in their docs: https://juliadiff.org/ReverseDiff.jl/dev/limits/

Not following this causes failures with using ReverseDiff:

```julia
@parameters σ ρ β
@variables x(t) y(t) z(t)

eqs = [D(D(x)) ~ σ * (y - x),
    D(y) ~ x * (ρ - z) - y,
    D(z) ~ x * y - β * z,
    w ~ x + y + z + 2 * β,
    ]

@mtkbuild sys = ODESystem(eqs, t)

ModelingToolkit.observed(sys)

u0 = [D(x) => 2.0,
    x => 1.0,
    y => 0.0,
    z => 0.0,]

p = [σ => 28.0,
    ρ => 10.0,
    β => 8 / 3,]

tspan = (0.0, 100.0)
prob = ODEProblem(sys, u0, tspan, p, jac = true)
sol = solve(prob, Tsit5())

mtkp = parameter_values(sol)
gt = rand(873)

gradient(mtkparams) do p
    sol = solve(prob, Rosenbrock23(), p = p, sensealg = SciMLSensitivity.GaussAdjoint(autojacvec = ReverseDiffVJP()))
    mean(abs.(sol[sys.x] .- gt))
end
```

cc @AayushSabharwal

### DIFF
--- a/src/array_partition.jl
+++ b/src/array_partition.jl
@@ -265,6 +265,8 @@ function Base._maybe_reshape(::IndexCartesian,
     Vector(A)
 end
 
+Base.IndexStyle(::ArrayPartition) = IndexLinear()
+
 ## recursive methods
 
 function recursivecopy!(A::ArrayPartition, B::ArrayPartition)


ChrisRackauckas:
But its index style is naturally Cartesian?

DhairyaLGandhi:
Ah, that is right, maybe we can restrict it to Vector of Vector? The reason is that when we call `canonicalize` we get an `ArrayPartition`, and that runs into this ReverseDiff limitation as I diagnosed this today with @AayushSabharwal. It runs into this assert https://github.com/JuliaDiff/ReverseDiff.jl/blob/bff801b75ea23ead2ece6f594ce1deee79898ba4/src/tracked.jl#L77

ChrisRackauckas:
Where does ReverseDiff come into play?

DhairyaLGandhi:
This is when we pass `sensealg = GaussAdjoint(autojacvec = ReverseDiffVJP())`

ChrisRackauckas:
With Array or ComponentArray? Is there a case that regresses/breaks or is it that case is not handled for the newer stuff?

DhairyaLGandhi:
It's not handled when we integrate SciMLStructures and  SII with Sensitivity.

PR #386, opened 2024-06-06
Fix copyto! overload for arraypartitions of scalars

ChrisRackauckas:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2182


ranocha:
Thanks for the fix! Could we please get a new release of RecursiveArrayTools.jl?

PR #398, opened 2024-08-10
Make sparsearrays into a weak dependency

ChrisRackauckas:


ext/RecursiveArrayToolsSparseArraysExt.jl
```diff
@@ -0,0 +1,20 @@
+module RecursiveArrayToolsSparseArraysExt
+
+import SparseArrays
+import RecursiveArrayTools
+
+function Base.copyto!(dest::SparseArrays.AbstractCompressedVector, A::RecursiveArrayToolsArrayPartition)
```
ChrisRackauckas, line 6:
```suggestion
function Base.copyto!(dest::SparseArrays.AbstractCompressedVector, A::RecursiveArrayTools.ArrayPartition)
```

PR #402, opened 2024-09-11
Fix similar for VectorOfArray

### LINKED ISSUE #401: Return `Vector` when creating similar `VectorOfArray` with new length
**Question❓**

Not sure if it's intended to behave like this or just a bug. When creating a similar array of `VectorOfArray` with a new length, wouldn't it be more intuitive to get `VectorOfArray` instead of just `Vector`? Since when we are doing the same for `Vector{Vector}`, we would end up getting `Vector{Vector}`.

```julia
julia> a = [rand(2) for _ in 1:3]
3-element Vector{Vector{Float64}}:
 [0.22672429007808392, 0.006545325869979246]
 [0.45202853665794596, 0.007153478061376006]
 [0.01950886804209273, 0.7556357057552318]

julia> va = VectorOfArray(a)
VectorOfArray{Float64,2}:
3-element Vector{Vector{Float64}}:
 [0.22672429007808392, 0.006545325869979246]
 [0.45202853665794596, 0.007153478061376006]
 [0.01950886804209273, 0.7556357057552318]

julia> similar(va, 4)
4-element Vector{Float64}:
 4.0e-323
 6.95219884198975e-310
 6.9522006963596e-310
 6.9521976152192e-310

julia> similar(a, 4)
4-element Vector{Vector{Float64}}:
 #undef
 #undef
 #undef
 #undef
```

ErikQQY:
Fix #401

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -29,6 +29,7 @@ returns a vector of the series for each component, that is, `A[i,:]` for each `i
 A plot recipe is provided, which plots the `A[i,:]` series.
 
 There is also support for `VectorOfArray` constructed from multi-dimensional arrays
+
 ```julia
 VectorOfArray(u::AbstractArray{AT}) where {T, N, AT <: AbstractArray{T, N}}
 ```
@@ -60,8 +61,9 @@ A[1, :]  # all time periods for f(t)
 A.t
 ```
 """
-mutable struct DiffEqArray{T, N, A, B, F, S, D <: Union{Nothing, ParameterTimeseriesCollection}} <:
-        AbstractDiffEqArray{T, N, A}
+mutable struct DiffEqArray{
+    T, N, A, B, F, S, D <: Union{Nothing, ParameterTimeseriesCollection}} <:
+               AbstractDiffEqArray{T, N, A}
     u::A # A <: AbstractVector{<: AbstractArray{T, N - 1}}
     t::B
     p::F
@@ -177,7 +179,9 @@ function DiffEqArray(vec::AbstractVector{T},
         ::NTuple{N, Int},
         p = nothing,
         sys = nothing; discretes = nothing) where {T, N}
-    DiffEqArray{eltype(T), N, typeof(vec), typeof(ts), typeof(p), typeof(sys), typeof(discretes)}(vec,
+    DiffEqArray{
+        eltype(T), N, typeof(vec), typeof(ts), typeof(p), typeof(sys), typeof(discretes)}(
+        vec,
         ts,
         p,
         sys,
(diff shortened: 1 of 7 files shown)


test/basic_indexing.jl
```diff
@@ -45,7 +45,7 @@ testvasim = similar(testva)
 @test size(testvasim) == size(testva)
 @test eltype(testvasim) == eltype(testva)
 testvasim = similar(testva, Float32)
-@test size(testvasim) == size(testva)
+#@test size(testvasim) == size(testva)
```
ErikQQY, line 48: `size` of a length-changed similar `VectorOfArray` would be broken since

```julia
julia> a = [rand(2) for _ in 1:3]
3-element Vector{Vector{Float64}}:
 [0.9768493678516263, 0.9216413916235213]
 [0.018985865691860893, 0.5682486867332623]
 [0.8908238990016157, 0.40453836312155733]

julia> similar(a, Vector{Float32}, 5)
5-element Vector{Vector{Float32}}:
 #undef
 #undef
 #undef
 #undef
 #undef
```

src/vector_of_array.jl
```diff
@@ -707,30 +717,37 @@ end
...
-        T, N, AT}) where {T, N, AT <: AbstractVector{<:AbstractArray{T}}}
-    return Base.similar(vec, eltype(vec))
+@inline function Base.similar(VA::VectorOfArray, ::Type{T} = eltype(VA)) where {T}
+    VectorOfArray([similar(VA[:, i], T) for i in eachindex(VA.u)])
 end
 
+@inline function Base.similar(VA::VectorOfArray, dims::N) where {N <: Number}
+    l = length(VA)
+    if dims <= l
+        VectorOfArray([similar(VA[:, i]) for i in 1:l])
```
ChrisRackauckas, line 741: This assumes the internal is a vector? Why not `VectorOfArray(similar.(VA.u))`?

ErikQQY: Yeah, broadcast is better

src/vector_of_array.jl
```diff
@@ -707,30 +715,37 @@ end
...
 Base.eltype(::Type{<:AbstractVectorOfArray{T}}) where {T} = T
-# TODO: Is there a better way to do this?
+
 @inline function Base.similar(VA::AbstractVectorOfArray, args...)
     if args[end] isa Type
-        return Base.similar(eltype(VA)[], args..., size(VA))
+        return Base.similar(VA.u, args..., size(VA))
     else
-        return Base.similar(eltype(VA)[], args...)
+        return Base.similar(VA.u, args...)
```
ChrisRackauckas, line 723: This won't rewrap.

ErikQQY: Just fixed

ChrisRackauckas (approved):

PR #419, opened 2025-01-28
Fix `fill!(::ArrayPartition)`

### LINKED ISSUE #418: Broadcasting on `ArrayPartition`s is not applied recursively
```
julia> x = CUDA.zeros(10_000_000);

julia> y = ArrayPartition(x);

julia> y .= 0
ERROR: Scalar indexing is disallowed.
Invocation of setindex! resulted in scalar indexing of a GPU array.
This is typically caused by calling an iterating implementation of a method.
Such implementations *do not* execute on the GPU, but very slowly on the CPU,
and therefore should be avoided.
```
This should recursively apply a broadcast to the elements of the partition, which should then correctly dispatch to the GPU code. This doesn't work, so some CPU code is called instead.

Edit: It looks like a simple
```julia
function Base.fill!(m::ArrayPartition, x)
    for i in eachindex(m.x)
        fill!(m.x[i], x)
    end

    return m
end
```
might do the trick.

efaulhaber:
Closes #418.


src/array_partition.jl
```diff
@@ -209,6 +209,13 @@ function Base.copyto!(A::ArrayPartition, src::ArrayPartition)
     A
 end
 
+function Base.fill!(A::ArrayPartition, x)
+    for i in eachindex(A.x)
+        fill!(A.x[i], x)
```
ChrisRackauckas, line 214: There's a lot of assumptions being made here. Is this `fill!` supposed to required that A.x are all the same size, and that it copies each one individually? Or is it supposed to match the total size of A and split x between the parts? I would've thought the latter since that is what matches the linear indexing.

efaulhaber: I don't understand. Do you mean that we could treat it like a vector of vectors and call fill with `x::Vector`?
```julia
julia> y = [[1, 2], [3, 4]]
2-element Vector{Vector{Int64}}:
 [1, 2]
 [3, 4]

julia> fill!(y, [5, 6])
2-element Vector{Vector{Int64}}:
 [5, 6]
 [5, 6]
```
This doesn't work on main either:
```julia
julia> x = ArrayPartition([1, 2], [3, 4])
([1, 2], [3, 4])

julia> fill!(x, [5, 6])
ERROR: MethodError: Cannot `convert` an object of type Vector{Int64} to an object of type Int64
The function `convert` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  convert(::Type{T}, ::Random123.__m128i) where T<:Union{Signed, Unsigned}
   @ Random123 /scratch/efaulha2/.julia/packages/Random123/d3ctu/src/x86/aesni_common.jl:16
  convert(::Type{T}, ::VectorizationBase.AbstractSIMD) where T<:Union{Bool, Float16, Float32, Float64, Int16, Int32, Int64, Int8, UInt16, UInt32, UInt64, UInt8, SIMDTypes.Bit}
   @ VectorizationBase /scratch/efaulha2/.julia/packages/VectorizationBase/wHnQd/src/base_defs.jl:201
  convert(::Type{T}, ::SentinelArrays.ChainedVectorIndex) where T<:Union{Signed, Unsigned}
   @ SentinelArrays /scratch/efaulha2/.julia/packages/SentinelArrays/vt8A6/src/chainedvector.jl:262
  ...

Stacktrace:
 [1] fill!(A::RecursiveArrayTools.ArrayPartition{Int64, Tuple{Vector{Int64}, Vector{Int64}}}, x::Vector{Int64})
   @ Base ./multidimensional.jl:1142
 [2] top-level scope
   @ REPL[15]:1
```
And it would be kinda awkward to use `fill!` like this, as it would require all `A.x` to have the same size.

efaulhaber: On main, `fill!(A::ArrayPartition, x)` will be forwarded to this:
```julia
function fill!(A::AbstractArray{T}, x) where T
    xT = convert(T, x)
    for I in eachindex(A)
        @inbounds A[I] = xT
    end
    A
end
```
Given how indexing and `setindex!` work on `ArrayPartition`s, this should be identical to my implementation for general arrays. I'm just calling the same function for each partition.
The only difference is that this can now be dispatched for GPU arrays.

ChrisRackauckas: Your implementation doesn't give the same answer as that because `fill!(A.x[i], x)` is filling the `i`th portion will all of `x`, where as the abstractarray dispatch is filling the ith portion with a sized subset of x.

efaulhaber: Could you give me an example where `x` is not scalar? The `AbstractArray` dispatch is writing `x` into each entry of `A`. And I write `x` into each entry of each partition of `A`. I still don't see how this is not the same.

ChrisRackauckas: oh wait, I was thinking `copyto` 😅

ChrisRackauckas:
Needs a test.

ChrisRackauckas (approved):

PR #425, opened 2025-03-14
Specialize `setindex!` for `VectorOfArray{StructArray}`

jlchan:
Addresses https://github.com/SciML/OrdinaryDiffEq.jl/issues/2625#issuecomment-2720056747

The issue is that `VectorOfArray` indexing will not mutate a `VectorOfArray{StructArray}` due to the way `StructArrays` treats assignment: https://juliaarrays.github.io/StructArrays.jl/stable/counterintuitive/#Modifying-a-field-of-a-struct-element. For example:
```julia
u = VectorOfArray(StructArray{MVector{1, Float64}}(ntuple(_ -> [1.0, 2.0], 1)))
u[1,1] = 100
@show u[1,1] # this is still 1.0
```


jlchan:
I don't think Downgrade and Format CI failures are related to this PR.

PR #426, opened 2025-03-14
Fix broadcast assignment for `VectorOfArray` with `StructArrays`

jlchan:
Attempt 2 to address https://github.com/SciML/OrdinaryDiffEq.jl/issues/2625#issuecomment-2720056747. I just applied the same approach taken in https://github.com/SciML/RecursiveArrayTools.jl/pull/425 for `setindex!` to broadcasted `Base.copyto!`.

I've verified that this also fixes the discrepancy in number of time-steps taken by an adaptive solver. The following MWE now results in the same number of accepted time-steps under this PR.
```julia
using OrdinaryDiffEqSSPRK, RecursiveArrayTools
using StaticArrays, StructArrays

function rhs!(du_voa, u_voa, params, t)
    du = parent(du_voa)
    u = parent(u_voa)
    du .= u
end

# StructArray storage
u = StructArray{SVector{1, Float64}}(ntuple(_ -> [1.0, 2.0], 1))
ode = ODEProblem(rhs!, VectorOfArray(u), (0, 0.7))
sol_SA = solve(ode, SSPRK43())
@show sol_SA.stats.naccept

# Vector{<:SVector} storage
u = SVector{1, Float64}.([1.0, 2.0])
ode = ODEProblem(rhs!, VectorOfArray(u), (0, 0.7))
sol_SV = solve(ode, SSPRK43())
@show sol_SV.stats.naccept
```

### DIFF
--- a/ext/RecursiveArrayToolsStructArraysExt.jl
+++ b/ext/RecursiveArrayToolsStructArraysExt.jl
@@ -3,7 +3,8 @@ module RecursiveArrayToolsStructArraysExt
 import RecursiveArrayTools, StructArrays
 RecursiveArrayTools.rewrap(::StructArrays.StructArray, u) = StructArrays.StructArray(u)
 
-using RecursiveArrayTools: VectorOfArray
+using RecursiveArrayTools: VectorOfArray, VectorOfArrayStyle, ArrayInterface, unpack_voa,
+                           narrays, StaticArraysCore
 using StructArrays: StructArray
 
 const VectorOfStructArray{T, N} = VectorOfArray{T, N, <:StructArray}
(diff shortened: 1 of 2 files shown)


jlchan:
@ChrisRackauckas do you know if the integration CI failures are spurious? The error messages don't appear to involve `StructArrays`, but I'm not sure if inference failures like [IntegrationTest / DiffEqBase.jl/Downstream/1 (pull_request)](https://github.com/SciML/RecursiveArrayTools.jl/actions/runs/13866614007/job/38806902300?pr=426) are related.

ChrisRackauckas:
Spurious, I need to work through those.

ranocha:
Could you please add some tests to OrdinaryDiffEq.jl to make sure that this does not regress again?

jlchan:
Sure, do you mean downstream tests in RecursiveArrayTools?

ranocha:
Or just regular tests in OrdinaryDiffEq.jl

jlchan:
How do I add StructArrays.jl to the list of OrdinaryDiffEq.jl test dependencies?

huiyuxie:
> Could you please add some tests to OrdinaryDiffEq.jl to make sure that this does not regress again?

I’m not exactly sure what it means, but if it’s something like running tests with a specific dependency, you can try something like here https://github.com/trixi-framework/Trixi.jl/blob/main/.github/workflows/downstream.yml)(https://github.com/trixi-framework/Trixi.jl/blob/main/.github/workflows/downstream.yml. In other words, treat StructArrays.jl as a downstream package, so you won’t have to add StructArrays.jl directly to OrdinaryDiffEq.jl, which would otherwise make it heavy.

ranocha:
You can add the package to the `[extras]` section and the `test = ...` assignment in the main Project.toml, e.g., https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/lib%2FOrdinaryDiffEqSSPRK%2FProject.toml

huiyuxie:
What is the difference between running tests through a Project.toml file and running them via a .github workflow? @ranocha

ranocha:
The tests will be run via a GitHub workflow, but the test dependencies need to be specified in a Project.toml file.

PR #428, opened 2025-03-27
Update ForwardDiff compat

oscardssmith:


oscardssmith:
@ChrisRackauckas these tests don't show anything useful (since OrdinaryDiffEq is in the test deps for RecursiveArrayTools). Should I just merge and hope?

PR #429, opened 2025-03-28
improve tests

oscardssmith:
This removes some deprecated indexing and the OrdinaryDiffEq dependency from the tests.


oscardssmith:
@ChrisRackauckas do you understand the LTS python segfault or the downstream test failures? They seem unrelated, but it's somewhat hard to say.

PR #433, opened 2025-04-07
Fix NamedArrayPartition similar when size is passed

SouthEndMusic:
@ChrisRackauckas can you make a release when this is merged?


ChrisRackauckas:
yes, it's still a draft though?

PR #434, opened 2025-04-21
Recurse adapt

ChrisRackauckas:
Needs a test case to really know if this is right. Thinking of GPUArrays, you can't have a GPUArray of GPUArrays so it should be an Array of GPUArrays


vchuravy:
Yeah Array of GPUArrays sounds right

PR #444, opened 2025-05-09
Type-generic zeromatrix for arraypartition

ChrisRackauckas:
One part of https://github.com/SciML/OrdinaryDiffEq.jl/issues/2703


ChrisRackauckas:
@oscardssmith this is blocked by a version-specific inference failure. Could I get some help with that? I assume it's because of some v1.10 vs v1.11 difference in mapreduce inference.

PR #447, opened 2025-05-15
Adds an abstract type to NamedArrayPartition

### LINKED ISSUE #443: Subtype ArrayPartition and NamedArrayPartition
We want to use a NamedArrayPartition as our u0 in a problem to replace a DEDataArray which we used (for not much good reason) and NamedArrayPartitions have the functionality we really want. However we need to be able to multiple dispatch on the 'type' of u0 we build (aka the fields we have assigned) and therefore want to use NamedArrayPartition as a supertype to subtype from. Something like the below:

```
mutable struct test_struct <: NamedArrayPartition
    x :: Vector{Real}
    y :: Matrix{Real}
    z :: AnyothersupportedtypebyNamedArrayPartition 
```

Snowd1n:
In order to resolve #443 where I would like to be able to subtype from NamedArrayPartition. This is a large restructuring of NamedArrayPartition but all tests within RecursiveArrayTools do pass. I'm not sure of other tests that I would need to run to ensure further compatibility. This maintains all current functionality but allows a user to create a new subtype of AbstractNamedArrayPartition as long as it maintains the same structure. I am pulling into this branch as this is where my previous fix to the ArrayInterface.zeromatrix lives.

## Checklist

## Additional context

### DIFF
--- a/src/named_array_partition.jl
+++ b/src/named_array_partition.jl
@@ -1,3 +1,5 @@
+abstract type AbstractNamedArrayPartition{T, A, NT} <: AbstractVector{T} end
+
 """
     NamedArrayPartition(; kwargs...)
     NamedArrayPartition(x::NamedTuple) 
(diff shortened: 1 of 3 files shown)


src/named_array_partition.jl
```diff
@@ -1,3 +1,5 @@
+abstract type AbstractNamedArrayPartition{T, A, NT} <: AbstractVector{T} end
```
ChrisRackauckas, line 1: I think this should have an AbstractArrayPartition, and that needs a docstring to describe its interface?

src/named_array_partition.jl
```diff
@@ -6,137 +8,155 @@ Similar to an `ArrayPartition` but the individual arrays can be accessed via the
...
 end
 
 # Note: overloading `getproperty` means we cannot access `NamedArrayPartition` 
 # fields except through `getfield` and accessor functions.
-ArrayPartition(x::NamedArrayPartition) = getfield(x, :array_partition)
+ArrayPartition(x::AbstractNamedArrayPartition) = getfield(x, :array_partition)
 
-function Base.similar(A::NamedArrayPartition)
-    NamedArrayPartition(
+# With new type structure this function does the same as Base.similar(x::AbstractNamedArrayPartition{T, S, NT}) where {T, S, NT}
```
ChrisRackauckas, line 35: Why is this one commented?

src/named_array_partition.jl
Snowd1n: I can add a higher level called AbstractArrayPartition{T} if you want and apply the same level of abstraction to the ArrayPartition interface as well (either or both) such that the same subtyping can be performed there also. I will add to the documentation, a docstring about how you can subtype from the AbstractNamedArrayPartition

ChrisRackauckas:
It's still missing what is meant by the interface. See for example https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/RecursiveArrayTools.jl#L16-L110. The questions that have to be answered are:

1. What does it mean to be an `AbstractNamedArrayPartition` / `AbstractArrayPartition`? What things must be true about the definition in order to satisfy the interface? For example, is there a `allowednames` function or something that everything should assume exists for `AbstractNamedArrayPartition`
2. What are the functions and allowed operations on any member of this group?
3. What is the relation to any other interface (for example `AbstractArray`)?

Then we need to make sure the abstract library implementations actually match this. For example, for `ArrayPartition`, the code currently always assumes `A.x`. See a chunk like https://github.com/SciML/RecursiveArrayTools.jl/blob/master/src/array_partition.jl#L127-L177. So there's a few decisions that have to be made here:

1. Is `ArrayPartition <: AbstractNamedArrayPartition` where the names are just `x`?
   * If that's the case, then we probably need to generalize those functions I linked to so that they match the interface, rather than being specific to ArrayPartition. Seems like a lot of code code could be deleted.
2. Are `AbstractNamedArrayPartition` expected to be non-allocating if they `.=`?
3. How is `AbstractNamedArrayPartition` expected to interact with LinearAlgebra?

PR #468, opened 2025-07-29
feat: Implement `recursivecopy` for structs

RomeoV:
Only supports mutable structs, and has to have extra logic to dispatch on basic types...

## Checklist

## Additional context

This problem came up as part of https://github.com/SciML/LinearSolve.jl/issues/648 and it a first draft of solving the issue. However, the current handling of the Julia special types is perhaps not perfect yet. Also the code path for normal types is now a bit more difficult to follow, and we explicitly don't allow copying mutable structs. Was that allowed before? In that case we can also just call `deepcopy` if it is mutable.


RomeoV:
I tried to swap this into [here](https://github.com/RomeoV/LinearSolve.jl/blob/ccad19062377bb07b53820bf77905acacdbd33b2/ext/LinearSolveForwardDiffExt.jl#L57) but that currently fails...

RomeoV:
I tried to swap this into [here](https://github.com/RomeoV/LinearSolve.jl/blob/ccad19062377bb07b53820bf77905acacdbd33b2/ext/LinearSolveForwardDiffExt.jl#L57) but that currently fails...

src/utils.jl
```diff
@@ -35,6 +49,42 @@ function recursivecopy(a::AbstractVectorOfArray)
...
+           T === Nothing || T === Missing || 
+           mod === Core || mod === Base
+end
+
+function recursivecopy(s::T) where {T}
+    # Only handle user-defined immutable structs. Many basic Julia types (String, Symbol, 
+    # Tuple, etc.) are technically structs but should use copy() or return as-is.
+    if Base.isstructtype(T) && !_is_basic_julia_type(T)
+        if Base.ismutabletype(T)
+            error("recursivecopy for mutable structs is not currently implemented. Use deepcopy instead.")
```
ChrisRackauckas, line 67: what's the issue there?

RomeoV:
I'm also becoming increasingly less certain that it's sensible to try to write a version that works for arbitrary struct compositions. Perhaps it would make more sense to specialize on a few types that we know may shows up in sciml solution structs.

PR #487, opened 2025-10-14
Implement `Base.resize!` for `ArrayPartition`s

efaulhaber:
Fixes https://github.com/SciML/OrdinaryDiffEq.jl/issues/2890.


ChrisRackauckas:
This looks right, but can you measure the invalidations?

PR #506, opened 2025-12-18
Add JET tests

JoshuaLampert:
## Checklist

## Additional context

As discussed in https://github.com/SciML/RecursiveArrayTools.jl/pull/501#issuecomment-3670062931, this adds JET.jl tests. I'll try to also fix any potential issues reported by JET.jl here.

### DIFF
--- a/test/jet_tests.jl
+++ b/test/jet_tests.jl
@@ -1,4 +1,4 @@
-using JET
+using JET, Test, RecursiveArrayTools
 
 # Get all reports first
 result = JET.report_package(RecursiveArrayTools; target_modules = (RecursiveArrayTools,))


.github/workflows/JET.yml
```diff
@@ -0,0 +1,60 @@
+name: JET
```
ChrisRackauckas: instead of a new workflow it should just be a new test group

JoshuaLampert: That won't be possible because we cannot add JET.jl to the Project.toml and test on julia pre-v1.12 and v1.12 due to package compat connstraints, see https://github.com/NumericalMathematics/DispersiveShallowWater.jl/pull/216#issuecomment-3218815131.

ChrisRackauckas: No it just needs to allow JET v0.9 and v0.11

ChrisRackauckas:
Oh actually that's just for implicit import checks. So this approach seems fine.

.github/workflows/JET.yml
JoshuaLampert: You mean something like [2acf709](https://github.com/SciML/RecursiveArrayTools.jl/pull/506/commits/2acf709dd2ceb73257ca2b1c5c6f102bd76b4a8e)? Let's see if that works. Last time I tried to make something like this work, I always got compat problems.

PR #526, opened 2026-01-19
Fix type instability for `vec[1, end]` indexing

### LINKED ISSUE #525: Type instability for `vec[1, end]`
**Describe the bug 🐞**

Accessing a recursive array with `[1, end]` is type-unstable. Was type stable on v3.41 and earlier, broken on 3.42 and later. Related to @JoshuaLampert 's recent work, it seems.

**Expected behavior**

Type stability for array accesses.

**Minimal Reproducible Example 👇**

```julia
vecvec = VectorOfArray([fill(2.0, 2) for i in 1:10])
access_end(vec) = vec[1, end]
@code_warntype access_end(vecvec)
```
yields
```
MethodInstance for access_end(::VectorOfArray{Float64, 2, Vector{Vector{Float64}}})
  from access_end(vec) @ Main REPL[4]:1
Arguments
  #self#::Core.Const(Main.access_end)
  vec::VectorOfArray{Float64, 2, Vector{Vector{Float64}}}
Body::Any
1 ─ %1 = Base.lastindex(vec, 2)::Core.PartialStruct(RecursiveArrayTools.RaggedEnd, Any[Core.Const(0), Int64])
│   %2 = Base.getindex(vec, 1, %1)::Any
└──      return %2
```

**Error & Stacktrace ⚠️**

Digging around with Cthulhu, I see:

Image
that is, starting at https://github.com/SciML/RecursiveArrayTools.jl/blob/6519ecf4fbf69a6134051d1fa11ad5685dd93c3a/src/vector_of_array.jl#L838,
`cols`, `prefix`, `resolved`, `inner_nd`, `padded` all are getting boxed somehow and killing the inference.

ChrisRackauckas-Claude:
## Summary
- Fixes #525
- Adds a specialized `getindex(A, i::Int, re::RaggedEnd)` method for type-stable access when using `end` indexing
- Adds a helper function `_ragged_getindex_int_col` as a function barrier for the Int column case
- Adds an early check in `_ragged_getindex` for the `RaggedEnd` sentinel case

## Problem
When accessing a `VectorOfArray` with `[1, end]`, the result was type-unstable (returning `Any`):
```julia
vecvec = VectorOfArray([fill(2.0, 2) for i in 1:10])
access_end(vec) = vec[1, end]
@code_warntype access_end(vecvec)  # Body::Any
```

The issue was that `lastindex(VA, d)` returns `RaggedEnd(0, offset)` as a sentinel for an already-resolved index, but the `getindex` dispatch couldn't infer the return type because `cols.dim == 0` is a runtime check.

## Solution
Add a specialized method for the common `vec[i, end]` pattern that handles the `RaggedEnd` type directly:
```julia
Base.@propagate_inbounds function Base.getindex(A::AbstractVectorOfArray, i::Int, re::RaggedEnd)
    if re.dim == 0
        return A.u[re.offset][i]
    else
        col = lastindex(A.u)
        resolved_idx = lastindex(A.u[col], re.dim) + re.offset
        return A.u[col][i, resolved_idx]
    end
end
```

After this fix:
```julia
@code_warntype access_end(vecvec)  # Body::Float64
```

## Test plan
- [x] Added test for type stability of `vec[1, end]` pattern
- [x] All existing tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/vector_of_array.jl
+++ b/src/vector_of_array.jl
@@ -576,6 +576,37 @@ function Base.:(:)(start::RaggedEnd, step::Integer, stop::Integer)
 end
 Base.broadcastable(x::RaggedRange) = Ref(x)
 
+# Specialized method for type stability when last index is RaggedEnd with dim=0 (resolved column index)
+# This handles the common case: vec[i, end] where end -> RaggedEnd(0, lastindex)
+Base.@propagate_inbounds function Base.getindex(
+        A::AbstractVectorOfArray, i::Int, re::RaggedEnd
+    )
+    if re.dim == 0
+        # Sentinel case: RaggedEnd(0, offset) means offset is the resolved column index
+        return A.u[re.offset][i]
+    else
+        # Non-sentinel case: resolve the ragged index for the last column
+        col = lastindex(A.u)
+        resolved_idx = lastindex(A.u[col], re.dim) + re.offset
+        return A.u[col][i, resolved_idx]
+    end
+end
+
+# Specialized method for type stability when first index is RaggedEnd (row dimension)
+# This handles the common case: vec[end, col] where end -> RaggedEnd(1, 0)
+Base.@propagate_inbounds function Base.getindex(
+        A::AbstractVectorOfArray, re::RaggedEnd, col::Int
+    )
+    if re.dim == 0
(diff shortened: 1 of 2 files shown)


JoshuaLampert (approved):
Thanks

PR #643, opened 2026-08-12
Implement the SciMLStructures interface for ArrayPartition

AJ0070:
## Checklist

## Additional context

- Implements the SciMLStructures interface for `ArrayPartition`, in `src/scimlstructures.jl` with SciMLStructures as a direct dependency.
- SciMLStructures opts every numeric array into the interface and canonicalizes it with `vec`. That is right for a contiguous array, but an `ArrayPartition` is already an `AbstractVector`, so `vec` returns it unchanged and the buffer comes back unflattened. The generic `repack` then errors, and its message asks for exactly this: "Please define the SciMLStructures interface for this type."
- Implements `canonicalize`, `replace` and `replace!` for the `Tunable` portion, laying the partitions out one after another so the flat buffer matches `collect(p)`.
- The buffer is `reduce(vcat, p.x)`. The partitions need not share an array type or an element type, so taking the buffer's type from one of them is wrong: `ArrayPartition([1, 2], [3.0, 4.5])` would give a `Vector{Int}` and truncate. Concatenating promotes across both, and keeps a uniformly static partitioning as an `MVector`.
- `reduce(vcat, (x,))` returns `x` itself, so a single partition is copied. Otherwise the buffer would alias `p` while `canonicalize` reports `aliases = false`. The test asserts this by mutating the buffer and checking the partitions are untouched.
- `replace` takes `new_values::AbstractArray` rather than leaving it untyped, which would be ambiguous with the generic `replace(::Tunable, ::AbstractArray, ::AbstractArray)`.
- Tests cover the flattening order, repack and replace agreeing, `replace!` writing into the same partition arrays rather than replacing them, uneven, single and empty partitions, mixed element and array types, integer element types, and the five non-tunable portions reporting absent.
- Motivated by SciML/LinearSolve.jl#1206, where a Krylov solve on an `ArrayPartition` right-hand side needs a flat buffer and a way back. Related: SciML/SciMLStructures.jl#79.
- Core test group passes locally.


AJ0070:
@ChrisRackauckas

ext/RecursiveArrayToolsSciMLStructuresExt.jl
```diff
@@ -0,0 +1,84 @@
...
+# `collect(p)`.
+
+# The buffer is a copy: the partitions are separate arrays, so no flat view over
+# them exists to alias. Built from the first partition rather than as a `Vector` so
+# that a GPU-backed partition yields a GPU-backed buffer.
+function _flatten(p::ArrayPartition)
+    isempty(p.x) && throw(
+        ArgumentError("cannot canonicalize an `ArrayPartition` with no partitions")
+    )
+    return copyto!(similar(first(p.x), length(p)), p)
```
ChrisRackauckas: It cannot assume that it all matches p.x[1] form.

ChrisRackauckas: `reduce(vcat, p.x)`?

AJ0070: addressed this

AJ0070:
@ChrisRackauckas

Project.toml
```diff
@@ -24,6 +24,7 @@ Mooncake = "da2b9cff-9c12-43a0-ae48-6db2b0edb7d6"
 MonteCarloMeasurements = "0987c9cc-fe09-11e8-30f0-b96dd679fdca"
 Polyester = "f517fe37-dbe3-4b94-8317-1923a5111588"
 ReverseDiff = "37e2e3b7-166d-5795-8a7a-e32c996b4267"
+SciMLStructures = "53ae85a6-f571-4167-b2af-e1d143709226"
```
ChrisRackauckas: just make a dep it's small

AJ0070:
@ChrisRackauckas both addressed.

`reduce(vcat, p.x)` for the flattening. It turned out to be a real bug rather than only a style point: sizing from `p.x[1]` gave a `Vector{Int}` for `ArrayPartition([1, 2], [3.0, 4.5])` and truncated the second partition. Concatenating promotes across differing element and array types, and keeps a uniformly static partitioning as an `MVector` rather than dropping it to a `Vector`. One wrinkle on top of it: `reduce(vcat, (x,))` returns `x` itself, so a single partition is copied, otherwise the buffer would alias `p` while `canonicalize` reports `aliases = false`.

SciMLStructures is now a direct dependency and the code moved from `ext/` to `src/scimlstructures.jl`.

Tests added for mixed element types, mixed array types and the single-partition aliasing, 53 assertions in total, Core group green locally. Description updated too, it still described the old buffer construction.

ChrisRackauckas (approved):

AJ0070:
@ChrisRackauckas sorry for the multiple pings. Fixed the Runic failure here, could you trigger CI again please?

PR #654, opened 2026-09-08
Fix VectorOfArray sum Int assertions on 32-bit Julia

ChrisRackauckas-Claude:
## Summary
- x86 CI fails two tests that assert `sum(...) === Int64(n)`
- On 32-bit, the sum is `Int32`; use `Int(n)` for native word size

## Test plan
- [ ] `tests / Core (…, x86)` green
- [ ] Existing x64 Core still green

Made with [Cursor](https://cursor.com)


ChrisRackauckas-Claude:
x86 Core is green on this PR. Merge when convenient — unblocks SciML 32-bit CI parity (no architecture-specific hard fails).

ChrisRackauckas-Claude:
x86 lane is green on this PR. Ready to merge when convenient (part of SciML 32-bit failure-parity sweep).

